PR #6, opened 2016-03-06
SBT fixes

pepegar:
Made two small tweaks to `build.sbt` file:

* Make `libraryDependecies` key more idiomatic. We dont need to append first an element and then a sequence.
* Change name from `uranio` to `fetch`.


dialelo:
Many thanks!

PR #17, opened 2016-05-06
Reorganize files and clean up warnings

raulraja:
The following PR cleans warnings and reorganizes files based on the Fetch model concerns isolatiing default implicit impls, so they can be overridden and imported properly.

@dialelo Can you please review? Thanks!


shared/src/main/scala/fetch.scala
```diff
@@ -0,0 +1,230 @@
...
+  }
+
+  implicit val fetchApplicative: Applicative[Fetch] = new Applicative[Fetch] {
+    def pure[A](a: A): Fetch[A] = Fetch.pure(a)
+
+    def ap[A, B](ff: Fetch[A => B])(fa: Fetch[A]): Fetch[B] =
+      Fetch.join(ff, fa).map({ case (f, a) => f(a) })
+  }
+
+  // Cache
```
dialelo: The comment got out of date.

dialelo:
Excellent job, besides the minor nitpick looks great to me :shipit:

PR #20, opened 2016-05-06
Abstracts over the target runtime Monad

raulraja:
The following PR decouples `DataSource` impls making them agnostic of the target runtime `M[_] : MonadError[A, Throwable]` so this only needs to be known to the actual call site run semantics.
As a side effect makes `Fetch` structures re-computable to any arbitrary `M[_]` such as `Future`, `Task` etc...

@dialelo Can you please review?


dialelo:
Outstanding!

PR #24, opened 2016-05-16
Documentation

dialelo:
@raulraja take a look and let me know your suggestions please.


docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,546 @@
...
+ - batching requests to the same data source
+ - requesting independent data from different sources in parallel
+ - caching previously seen results
+
+However, if we mix these optimizations with the code that fetches the data
+we may end up trading clarity for performance. Furthermore, we are
+mixing low-level (optimization) and high-level (business logic with the data
+we read) concerns.
+
+Fetch is a library that allows your data fetches to be written in a concise,
```
raulraja: This should be the first paragraph

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,546 @@
...
+
+# Installation
+
+First of all, add the following dependency to your SBT build file:
+
+```scala
+"org.fortysevendeg" %%% "fetch" % "0.1.0"
+```
+
+Now you'll have Fetch available both in Scala and Scala.js.
```
raulraja: We should provide both imports separately indicating the targets

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,546 @@
...
+  def fetch(ids: List[Identity]): Eval[Map[Identity, Result]]
+}
+```
+
+It takes two type parameters:
+
+ - `Identity`: the identity we want to fetch (a `UserId` if we were fetching users)
+ - `Result`: the type of the data we retrieve (a `User` if we were fetching users)
+
+The `fetch` method takes a non empty list of identities and must return an `Eval` that will result
```
raulraja: We should link Cats types to their docs

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,546 @@
...
+couldn't be fetched or no longer exists.
+
+Returning `Eval` makes it possible to defer evaluation with a concurrency monad when running a fetch. For example
+when using `Future` as your target monad when running a fetch the data source fetches will possibly run in a thread pool,
+depending on the implicit execution context you use.
+
+## Writing your first data source
+
+Now that we know about the `DataSource` typeclass, let's write our first data source! We'll be using Scala's
+`Future` as our concurrency monad for the examples. We'll write a little function for simulating latency and
```
raulraja: This is going to be confusing for users since they won't now this is not part of the actual runtime but a test function. IMO it does not belong in the docs specially not when they have not even seen the lib in action.

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,546 @@
...
+Now that we have a data source we can write a function for fetching users
+given an id, we just have to pass a `UserId` as an argument to `Fetch`.
+
+```scala
+def getUser(id: UserId): Fetch[User] = Fetch(id) // or, more explicitly: Fetch(id)(UserSource)
+```
+
+## Declaring and running a fetch
+
+We are now ready to declare and run fetches, the only thing missing is an instance
```
raulraja: This is nice but probably belongs in an advance section of the docs when you want to create your custom runtimes. Since we are gonna provide implicits for most runtimes Task, Future, etc... this is only a special case for those implementing their own.

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,546 @@
...
+
+### Batching
+
+If we combine two independent requests to the same data source, Fetch will
+automatically batch them together into a single request. Fetch's `join` operator
+help us tell the library that two fetches are independent, and thus can be batched
+if they use the same data source:
+
+```scala
+val fch: Fetch[(User, User)] = Fetch.join(getUser(1), getUser(2))
```
raulraja: I think the examples should use the syntax instead as it is more user friendly than the more verbose companion object methods.

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,546 @@
...
+of data very easy.
+
+### Parallelism
+
+Combining multiple independent requests to the same data source results in batching but what happens
+when combining independent requests to *different* data sources? We have the oportunity to query the
+different data sources at the same time.
+
+Since we are using `Future` as our concurrency monad, the requests will effectively run in parallel,
+each in its own thread.
```
raulraja: logical thread

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,546 @@
...
+//=> (Post(1,2,An article with id 1),User(2,@egg_2))
+```
+
+As you can notice, both requests are started at the same time in different threads (with ids 83 and 84) and
+the fetch finishes as soon as both results are available.
+
+## Combinators
+
+Besides `flatMap` for sequencing fetches and `join` for running them concurrently, Fetch provides a number of
+other combinators that you can use.
```
raulraja: We can get rid of `that you can use` here

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,546 @@
...
+
+### Collect
```
raulraja: Is order respected in this combinator? Also collect may be a misleading name because of the existing `collect` combinator for `map + filter` using Partial functions predicates

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,546 @@
...
+
+# Error handling
+
+You probably have noticed that `DataSource.fetch` takes a list of identities and returns a map of identities to their result, taking
+into account the possibility of some identities not being found. Whenever an identity isn't found, the fetch execution will
+fail.
+
+## Diagnosing fetch failures
+
+Whenever a fetch fails, a `FetchFailure` exception is thrown. The `FetchFailure` will have the environment, which gives you information about the execution of a fetch.
```
raulraja: I'm guessing the exception is not actually thrown but captured by the MonadError instance which depending on your target may throw it like in the case of `Id` but other better fit ones like `Task` would place it on the Left.

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,546 @@
...
+// <~~~[thread: 56] Fetching users List(1, 2)
+
+//=> @egg_1 is friends with @egg_2
+```
+
+### Traverse
+
+Apart from using Fetch with cartesian syntax and performing all fetches in a product concurrently, we can treat Fetch as an applicative and get implicit concurrency and batching as well.
+
+For the next example, we'll import cats' extensions to the `List` type and traverse syntax. This will allow us to use both `sequence` and `traverse` methods on lists, which are equivalent to `Fetch#collect` and `Fetch#traverse`.
```
raulraja: So `Fetch#collect` should be `Fetch#sequence`?

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,546 @@
...
+await(fut)
+// ~~~>[thread: 62] Fetching users List(1, 2, 3)
+// <~~~[thread: 62] Fetching users List(1, 2, 3)
+
+//=>  List(User(1,@egg_1), User(2,@egg_2), User(3,@egg_3))
+```
+
+## scalaz
+
+There is still no integration with Scalaz, but soon will be.
```
raulraja: We can remove this until ready

docs/src/jekyll/index.md
```diff
@@ -0,0 +1,170 @@
...
+ - cats: ["Cats", "Fetch uses cats' Free Monad implementation as well as some of its data types."]
+ - fp: ["Functional Programming", "Fetch is implemented using the Free Monad and Interpreter pattern."]
+---
+
+## Installation
+
+Add the following dependency to your project's build file.
+
+```scala
+"org.fortysevendeg" %%% "fetch" %% "0.1.0"
```
raulraja: This should be "com.fortysevendeg". also I recall there was an internal proposal to change our internal packages to something shorter.

docs/src/jekyll/docs.md
dialelo: Unfortunately cats' `Eval` doesn't have docs yet, which is a shame. I'm going to take some time to contribute documentation for `Eval` and will add links when it's merged.

docs/src/jekyll/docs.md
dialelo: Yes, should we use `sequence` instead?

docs/src/jekyll/docs.md
dialelo: The dependency declaration with `%%%` should cover both Scala and Scala.js but i'll include all possibilities: scala only, scala.js only and both.

docs/src/jekyll/docs.md
dialelo: Yep, going to change that.

dialelo:
@raulraja i've changed many examples to use `Id` instead of `Future`, I'm going to do so with the homepage to simplify it. Let me know what you think!

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,562 @@
...
+
+```scala
+def getUser(id: UserId): Fetch[User] = Fetch(id) // or, more explicitly: Fetch(id)(UserSource)
+```
+
+## Declaring and running a fetch
+
+We are now ready to declare and run fetches. We need to provide Fetch with a target
+monad when we want to execute a fetch. We'll be using `Id` for now, make sure to import
+`fetch.implicits._` since Fetch needs an instance of `MonadError[Id, Throwable]` for running
```
raulraja: Also for any other Monads. We should state right here that Fetch comes with builtin impls of MonadError for Id, Task, Future, etc...

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,562 @@
...
+
+- one for retrieving a post given a post id
+- another for retrieveng post metadata given a post id
+
+```scala
+implicit object PostSource extends DataSource[PostId, Post]{
+  override def fetch(ids: List[PostId]): Eval[Map[PostId, Post]] = {
+    Eval.later({
+      println(s"Fetching posts $ids")
+      ids.map(i => (i, Post(i, i % 3 + 1, s"An article with id $i"))).toMap
```
raulraja: I think this use of map to generate data is confusing. I think each datasource should have a list, vector or map call Database or whatever that illustrates that the data may be elsewhere and the Datasource is nothing but an access layer on top of it.

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,562 @@
...
+//=> result: (Post, User) = (Post(1,2,An article with id 1),User(2,@egg_2))
+```
+
+In the previous example we fetched a post given its id, and then fetched its author. These
+data could come from entirely different places but Fetch makes working with heterogeneous sources
+of data very easy.
+
+### Parallelism
+
+Combining multiple independent requests to the same data source results in batching but what happens
```
raulraja: Kind of a run-on sentence should probably be split in two for clarity

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,562 @@
...
+val fch: Fetch[(Post, User)] = getPost(1).product(getUser(2))
+
+val result: (Post, User) = Fetch.run[Id](fch)
+// Fetching posts List(1)
+// Fetching users List(2)
+
+//=> result: (Post, User) = (Post(1,2,An article with id 1),User(2,@egg_2))
+```
+
+Since we are interpreting the fetch to the `Id` monad, that doesn't give us any concurrency, the fetches
```
raulraja: We may want to describe here or somewhere more appropiate that Fetches are reusable and can be interpreted to different runtimes for example Task or Future in production code and Id in tests

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,562 @@
...
+will be run sequentially. However, if we interpret it to a `Future` each request will run in its own logical
+thread.
+
+## Combinators
+
+Besides `flatMap` for sequencing fetches and `product` for running them concurrently, Fetch provides a number of
+other combinators.
+
+### Sequence
```
raulraja: lowercase s and backsticks? `sequence`

raulraja: nevermind, it's fine :)

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,598 @@
...
+```scala
+val fch: Fetch[List[User]] = List(1, 2, 3).traverse(getUser)
+
+val result: List[User] = Fetch.run[Id](fch)
+// Fetching users List(1, 2, 3)
+
+//=> result: List[User] = List(User(1,@egg_1), User(2,@egg_2), User(3,@egg_3))
+```
+
+# Interpreting a fetch to a concurrency monad
```
raulraja: async capable monad

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,598 @@
...
+
+Albeit the examples use `Id` as the concurrency Monad, `Fetch` is not limited to just `Id`, any monad `M` that
```
raulraja: ...any type constructor `M[_]` that implements MonadError...

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,598 @@
...
+implements `MonadError[M, Throwable]` will do. Fetch provides `MonadError` instances for some existing monads like
+`Future`, `cats.Id` and `cats.Eval` and is easy to write your own.
+
+In practice you'll be interpreting a fetch to a concurrency monad like `Future` or `scalaz.concurrent.Task` to exploit
```
raulraja: If we namespace Task we should namespace Future too as scalaz.concurrent also includes a lower primitive type called `Future` which is contained inside `Task`

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,598 @@
...
+import scala.concurrent.duration._
+import scala.concurrent.ExecutionContext.Implicits.global
+
+val fch: Fetch[(User, Post)] = (getUser(1) |@| getPost(1)).tupled
+
+val fut: Future[(User, Post)] = Fetch.run[Future](fch)
+// Fetching users List(1)
+// Fetching posts List(1)
+
+Await.result(fut, 1 seconds)
```
raulraja: Add comment //blocking call for illustration purposes, Don't try this at home!

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,598 @@
...
+in the cache.
+
+If only part of the data is cached, the cached data won't be asked for:
+
+```scala
+val fch: Fetch[List[User]] = List(1, 2, 3).traverse(getUser)
+val cache = InMemoryCache(UserSource.identity(2) -> User(2, "@two"))
+
+val result: List[User] = Fetch.run[Id](fch, cache)
+// Fetching users List(1, 3)
```
raulraja: All of these comments are pointless and so is the println when we have tut implemented since you can tell tut to display the REPL result for a given snippet selectively.

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,598 @@
...
+
+//=> result: List[User] = List(User(1,@egg_1), User(2,@two), User(3,@egg_3))
+```
+
+## Replaying a fetch
```
raulraja: This is misleading because Fetches can be replayed regardless of caching but this section talks more about how you can reuse a Fetch cache

docs/src/jekyll/docs.md
dialelo: Added it just below this paragraph.

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,598 @@
...
+val env: Env = Fetch.runEnv[Id](fch)
+// Fetching users List(1, 2, 3)
+
+val result: List[User] = Fetch.run[Id](fch, env.cache)
+//=> result: List[User] = List(User(1,@egg_1), User(2,@egg_2), User(3,@egg_3))
+```
+
+## Implementing a custom cache
+
+The default cache is implemented as an in-memory map, but users are free to use their own caches when running a fetch. Your cache should implement the `DataSourceCache` trait, and after that you can pass it to Fetch's `run` methods.
```
raulraja: Perhaps worth mentioning here that the cache is trampolined inside a state monad when evaluating a Fetch so people understand they can easily implement caches with inmutable data structures and not mutable maps.

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,598 @@
...
+//=> result: User = User(1,dialelo)
+```
+
+# Error handling
+
+As we mentioned before, when interpreting a fetch to a target monad `M`, an implicit instance of `MonadError[M, Throwable]` has to be
+available. [`MonadError`](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/MonadError.scala) gives us a few combinators
+for working with errors, like `MonadError#raiseError` and `MonadError#attempt`.
+
+One of the most interesting combinators is `attempt`, which given a `M[A]` yields a `M[Xor[Throwable, A]]`. Knowing this, we can run fetches
```
raulraja: Let's preffer more idiomatic versions such as `M[Throwable Xor A]`

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,598 @@
...
+## Missing identities
+
+You probably have noticed that `DataSource.fetch` takes a list of identities and returns a map of identities to their result, taking
+into account the possibility of some identities not being found. Whenever an identity isn't found, the fetch execution will
+fail.
+
+Whenever a fetch fails, a `FetchFailure` exception is thrown. The `FetchFailure` will have the environment, which gives you information
+about the execution of the fetch.
+
+# Syntax
```
raulraja: Awesomesauce!

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,598 @@
...
+
+val result: List[User] = Fetch.run[Id](fch)
+// Fetching users List(1, 2, 3)
+
+//=> result: List[User] = List(User(1,@egg_1), User(2,@egg_2), User(3,@egg_3))
+```
+
+## cats
+
+Fetch is built using cats' Free monad construction and thus works out of the box with
```
raulraja: Mention to tree optimization?

docs/src/jekyll/docs.md
```diff
@@ -0,0 +1,598 @@
...
+cats syntax. Using cats' syntax we can make fetch declarations more concise, without
+the need to use the combinators in the `Fetch` companion object.
+
+### Apply
```
raulraja: I'd just say here Applicative instead of Apply

raulraja:
Some small comments, once addressed this LGTM and it's ready for lang review. Awesome work!

docs/src/jekyll/docs.md
dialelo: I'll change the title to "Replaying a fetch without querying any data source" then.

docs/src/jekyll/docs.md
dialelo: Done!

dialelo:
All your comments addressed, feel free to merge @raulraja :sparkles:

raulraja:
Great work @dialelo. 
@MaureenElsberry This is ready for lang review. Can you take a quick pass? thanks!

PR #27, opened 2016-05-16
remove unused type parameter

lambdista:


dialelo:
Good catch! Will merge as soon as Travis is happy. Thanks for your time.

lambdista:
Thank you for your work!

PR #33, opened 2016-05-20
RR Fetch Syntax

raulraja:
The following PR provides implicit syntax to lift any `A` to ` Fetch[A]` as well as syntax for the most useful combinators on a `Fetch[A]`.

@dialelo @rafaparadela Can you please review? Thanks!

Pseudo extract from the included docs contribution:
# Syntax
## Implicit syntax
Fetch provides implicit syntax to lift any value to the context of a `Fetch` in addition to the most common used combinators active within `Fetch` instances.
### pure
Plain values can be lifted to the Fetch monad with `value.fetch`:
```scala
val fetchPure: Fetch[Int] = 42.fetch
```
Executing a pure fetch doesn't query any data source, as expected.
```scala
val result: Int = fetchPure.runA[Eval].value
```
### error
Errors can also be lifted to the Fetch monad via `exception.fetch`. Note that interpreting
an errorful fetch to `Eval` will throw the exception so we won't do that:
```scala
val fetchFail: Fetch[Int] = (new Exception("Something went terribly wrong")).fetch[Int]
```
### join
We can compose two independent fetches with `fetch1.join(fetch2)`.
```scala
val fetchJoined: Fetch[(Post, User)] = getPost(1).join(getUser(2))
```
If the fetches are to the same data source they will be batched; if they aren't, they will be evaluated at the same time.
```scala
val result: (Post, User) = fetchJoined.runA[Eval].value
```
### runA
Run directly any fetch to a target any target `Monad` with a `MonadError` instance in scope `fetch1.runA[Eval]`.
```scala
val post: Eval[Post] = getPost(1).runA[Eval]
```
### runE
Extract a fetch an get it's runtime environment `fetch1.runE[Eval]`.
```scala
val env: Eval[FetchEnv] = getPost(1).runE[Eval]
```
### runF
Run a fetch obtaining the environment and final value `fetch1.runF[Eval]`.
```scala
val env: Eval[(FetchEnv, Post)] = getPost(1).runF[Eval]
```

### DIFF
--- a/docs/src/tut/docs.md
+++ b/docs/src/tut/docs.md
@@ -615,10 +615,17 @@ val result: Int = Fetch.run[Eval](fetchPure).value
 ### error
 
 Errors can also be lifted to the Fetch monad, in this case with `Fetch#error`. Note that interpreting
-an errorful fetch to `Eval` will throw the exception so we won't do that:
+an errorful fetch to `Eval` won't throw the exception unless we access the value with the `.value` method.
 
-```tut:silent
+A safer way to deal with errors is to use MonadError's `attempt` to turn the exception into a `Xor.Left` value:
+
+```tut:book
+val ME = implicitly[MonadError[Eval, Throwable]]
+
 val fetchFail: Fetch[Int] = Fetch.error(new Exception("Something went terribly wrong"))
+val result: Eval[Int] = fetchFail.runA[Eval]
+val safeResult: Eval[Throwable Xor Int] = ME.attempt(result)
+val finalValue: Throwable Xor Int = safeResult.value
 ```
 
 ### join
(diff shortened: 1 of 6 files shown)


docs/src/tut/docs.md
```diff
@@ -522,9 +524,75 @@ about the execution of the fetch.
...
+Executing a pure fetch doesn't query any data source, as expected.
+
+```tut:book
+val result: Int = fetchPure.runA[Eval].value
+```
+
+### error
+
+Errors can also be lifted to the Fetch monad via `exception.fetch`. Note that interpreting
+an errorful fetch to `Eval` will throw the exception so we won't do that:
```
dialelo: In the case of `Eval` it won't throw the exception unless accessing the value with `.value`, maybe we can show how the exception is conveyed on the `Xor.Left` when ussing the `attempt` combinator?

dialelo:
Other than the comment LGTM, really awesome work :+1:

raulraja:
Feel free to reword or remove that part for now if you want this to be included in the release.I won't be close to a computer this evening

dialelo:
All right, i'll edit that part and cut another release with the new syntax and docs.

dialelo:
:shipit:

PR #34, opened 2016-05-20
Fix error reporting on concurrent fetch failures

dialelo:
Makes sure that all the identities have been fetched in a concurrent round. @raulraja please take a look, I want to cut a bugfix release with this.


raulraja:
LGTM!

dialelo:
:shipit:

Issue #38, opened 2016-05-21
Make README.md a mirror of the documentation


raulraja:
The current README just points to the docs with a link but users already on Github could careless about jumping extra hoops. Also the file won't be indexed by Scaladex or crawlers that use the README as the doc defacto standard. Ideally post tut processing the file copied inside the jekyll site would also replace the README.md at the root and will auto generate an index to be placed a top the file to jump to sections.

PR #39, opened 2016-05-22
Typechecked and more thorough README

dialelo:


dialelo:
@raulraja take a look, addresses some of your points in #38. I'm not a big fan of putting al the docs in the README but I think showing what the library can do is enough, users that want more information will go to the more thorough documentation site.

Do you think a table of contents is needed?

dialelo:
Forgot to mention, see [the end result](https://github.com/47deg/fetch/blob/better-readme/README.md)

raulraja:
Looks great and no table of contents is needed. I think this covers those user that may run into the lib without a specific need but are curious to see what it does. If there is new stuff we may want to have Maureen do a quick Lang pass. LGTM!

dialelo:
No new stuff yet except for the change in fetch's signature, going to release 0.2.0 with this and the new syntax you contributed! :shipit:

Issue #42, opened 2016-05-25
Acknowledgments 


fwbrasil:
I'm not sure I'm correct, but this lib seems to be inspired by similar projects like Clump and Stitch. Could you please add an "Acknowledgments" section to the documentation mentioning it?

BTW, the generic approach for any monad and the typeclasses for sources are very nice! 👏 

Thanks! :)

dialelo:
I mentioned Clump, Stich, Haxl and few others in my talk and thought about adding acknowledgements/alternatives to the documentation but forgot to do it, thanks a lot for mentioning it! I owe a lot to all those projects, standing on the shoulders of giants here :smile: 

Thanks for the kind words!

PR #43, opened 2016-05-25
Add alternatives and acknowledgements

dialelo:
Addresses #42, @fwbrasil let me know if i am missing something please. And sorry for not including this earlier.

### DIFF
--- a/docs/src/tut/docs.md
+++ b/docs/src/tut/docs.md
@@ -50,6 +50,15 @@ def println(msg: String): Unit = {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


fwbrasil:
@dialelo that was fast, thanks! 👍

dialelo:
@raulraja take a look please!

raulraja:
LGTM. Not sure if it makes sense here but perhaps we should consider adding a matrix with the differences and also license types. In any case this is good to go.

PR #50, opened 2016-05-26
Don't depend on the cache contents for simplifying joined trees

### LINKED ISSUE #47: Don't depend on the cache contents for simplifying joined trees

dialelo:
@raulraja please take a look. 

This opens the door to using custom caches that don't remember all the elements, for instance caches with an LRU policy. Also if we where using a custom remote cache (like memcached) that is unreachable and thus we cannot read from it (return `None` from `DataSourceCache#get`), fetches will continue without problem querying the data source.

Previously wasn't possible to use a custom cache that forgot elements so I'd like to cut a release with these changes.

Closes #47

### DIFF
--- a/shared/src/main/scala/fetch.scala
+++ b/shared/src/main/scala/fetch.scala
@@ -34,7 +34,7 @@ final case class Cached[A](a: A)                            extends FetchOp[A]
 final case class FetchOne[I, A](a: I, ds: DataSource[I, A]) extends FetchOp[A]
 final case class FetchMany[I, A](as: NonEmptyList[I], ds: DataSource[I, A])
     extends FetchOp[List[A]]
-final case class Concurrent(as: List[FetchMany[_, _]]) extends FetchOp[Env]
+final case class Concurrent(as: List[FetchMany[_, _]]) extends FetchOp[DataSourceCache]
 final case class FetchError[A, E <: Throwable](err: E) extends FetchOp[A]
 
 object `package` {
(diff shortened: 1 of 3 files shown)


shared/src/main/scala/fetch.scala
```diff
@@ -148,52 +151,53 @@ object `package` {
...
+              fetched.fold(many: FetchOp[B])(results => Cached(results))
+            }
+          case conc @ Concurrent(manies) => {
+              val newManies = manies
+                .filterNot({ fm =>
+                  val ids: NonEmptyList[Any] = fm.as.asInstanceOf[NonEmptyList[Any]]
+                  val ds: DataSource[Any, _] = fm.ds.asInstanceOf[DataSource[Any, _]]
+
+                  ids
+                    .map(id => {
```
raulraja: Nothing wrong with this but the I'd and block it's verbose. May be simplified like the next chained call

shared/src/test/scala/FetchTests.scala
```diff
@@ -638,6 +638,44 @@ class FetchTests extends FreeSpec with Matchers {
 
     totalFetched(rounds) shouldEqual 0
   }
+
+  case class ForgetfulCache() extends DataSourceCache {
```
raulraja: final may be better suited since no point of ever extending this case class

raulraja:
LGTM!

shared/src/main/scala/fetch.scala
dialelo: I've refactored this in another branch, will submit a PR soon.

shared/src/test/scala/FetchTests.scala
dialelo: Compiler throws a warning when marking inner case classes as final, i'll leave it for now.

PR #51, opened 2016-05-26
Fix feature warning, typo, tab

williamho:
The usage of `1 seconds` causes a warning to appear [in the docs](http://47deg.github.io/fetch/docs#interpreting-a-fetch-to-an-async-capable-monad-3-future-0):

```scala
// warning: there was one feature warning; re-run with -feature for details
```


dialelo:
Good catch, thanks a lot!

PR #53, opened 2016-05-27
Separate fetching a single identity and fetching many in DataSource

dialelo:
@raulraja please take a look. @MaureenElsberry if you could do a language pass to the edited docs would be great.

### DIFF
--- a/docs/src/tut/docs.md
+++ b/docs/src/tut/docs.md
@@ -76,16 +76,15 @@ It takes two type parameters:
  - `Result`: the type of the data we retrieve (a `User` if we were fetching users)
 
 There are two methods: `fetchOne` and `fetchMany`. `fetchOne` receives one identity and must return
-an [Eval](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Eval.scala) that will result
-in an optional result. Returning an `Option` Fetch can detect whether an identity couldn't be fetched or no longer exists.
+an [Eval](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Eval.scala) containing
+an optional result. Returning an `Option` Fetch can detect whether an identity couldn't be fetched or no longer exists.
 
-`fetchMany` method takes a non-empty list of identities and must return an `Eval`  that will result
-in a map from identities to results. Accepting a list of identities gives Fetch the ability to batch requests to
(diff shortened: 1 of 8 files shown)


docs/src/tut/docs.md
```diff
@@ -123,6 +134,23 @@ given an id, we just have to pass a `UserId` as an argument to `Fetch`.
 def getUser(id: UserId): Fetch[User] = Fetch(id) // or, more explicitly: Fetch(id)(UserSource)
 ```
 
+
+### Data sources that don't support batcihng
```
raulraja: Batching

shared/src/main/scala/datasource.scala
```diff
@@ -19,12 +19,48 @@ package fetch
...
+    */
+  def fetchMany(ids: NonEmptyList[I]): Eval[Map[I, A]]
+
+  /** Use `fetchOne` for implementing of `fetchMany`. Use only when the data
+    * source doesn't support batching.
+    */
+  def batchingNotSupported(ids: NonEmptyList[I]): Eval[Map[I, A]] = {
+    val idsList = ids.unwrap
+    idsList
+      .map(fetchOne)
```
raulraja: When we use Task here we can paralelize this with a strategy argument

shared/src/test/scala/FetchTests.scala
```diff
@@ -34,31 +34,39 @@ object TestHelper {
...
   implicit object ManySource extends DataSource[Many, List[Int]] {
     override def name = "ManySource"
-    override def fetch(ids: NonEmptyList[Many]): Eval[Map[Many, List[Int]]] =
+    override def fetchOne(id: Many): Eval[Option[List[Int]]] =
+      M.pure(Option(0 until id.n toList))
+    override def fetchMany(ids: NonEmptyList[Many]): Eval[Map[Many, List[Int]]] =
       M.pure(ids.unwrap.map(m => (m, 0 until m.n toList)).toMap)
   }
 
   case class Never()
```
raulraja: We should make all case classes final

raulraja:
LGTM!

shared/src/main/scala/datasource.scala
dialelo: Working on that on a separate branch, will take this into account!

Issue #52, opened 2016-05-27
Support data sources that can only be queried asynchronously


dialelo:
Since currently data sources assume you can return a computation that synchronously gives you the result, the usefulness of the library is very limited in Scala.js. It would be desirable to support both synchronous and asynchronous data sources, maybe using `monix-eval`'s `Task` type instead of `Eval` in data source's fetch methods? `Task` is more general than `Eval`, although we'd lose the ability to run fetches synchronously.

PR #54, opened 2016-05-29
Support asynchronous data sources

dialelo:
Addresses #52. I also made an improvement: calling `fetchOne` and `fetchMany` depending on the number of the identities to be fetched. @raulraja please take a look!

- [x] Use `Id` as the target monad in the examples
- [x] Document all the available monad targets
- [x] Document `Query` constructors
- [x] Document how to bring your own concurrency monad
- [x] Simplify `Query` constructors: only leave `sync` (receiving a thunk), `async` and `eval` (lifts eval to synchronous query) 

Edit: leaving more thorough testing as part of #32

~~- [ ] Test Monix integration more thoroughly~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,45 +89,23 @@ import fetch.syntax._
 val fetchOne: Fetch[String] = fetchString(1)
 ```
 
-Now that we have created a fetch, we can run it to a `Task`. Note that when we create a task we are not computing any value yet. Having a `Task` instance allows us to try to run it synchronously or asynchronously, choosing a scheduler.
+We'll run our fetches to the ambiend `Id` monad in our examples, let's do some imports.
 
 ```scala
-import fetch.implicits._
-// import fetch.implicits._
-
-import scala.concurrent._
-// import scala.concurrent._
-
-import ExecutionContext.Implicits.global
-// import ExecutionContext.Implicits.global
-
-val result: Future[String] = fetchOne.runA[Future]
-// result: scala.concurrent.Future[String] = List()
+import cats.Id
+import fetch.unsafe.implicits._
+import fetch.syntax._
 ```
 
-Since we calculated the results eagerly using `Task#now`, we can run this fetch synchronously.
+Note that in real-life scenarios you'll want to run a fetch to a concurrency monad, synchronous execution of a fetch
+is only supported in Scala and not Scala.js and is meant for experimentation purposes.
+
+Let's run it and wait for the fetch to complete:
 
 ```scala
-import scala.concurrent.duration._
-// [152] One ToString 1
-// import scala.concurrent.duration._
-
-Await.result(result, Duration.Inf)
-// res3: String = 1
-```
-
-As you can see in the previous example, the `ToStringSource` is queried once to get the value of 1.
-
-
-```scala
-import scala.concurrent._
-// import scala.concurrent._
-
-import scala.concurrent.duration._
-// import scala.concurrent.duration._
-
-def await[A](t: Future[A]): A = Await.result(t, Duration.Inf)
-// await: [A](t: scala.concurrent.Future[A])A
+fetchOne.runA[Id]
+// [102] One ToString 1
+// res3: cats.Id[String] = 1
 ```
 
 ## Batching
@@ -139,19 +117,15 @@ import cats.syntax.cartesian._
 // import cats.syntax.cartesian._
 
 val fetchThree: Fetch[(String, String, String)] = (fetchString(1) |@| fetchString(2) |@| fetchString(3)).tupled
-// fetchThree: fetch.Fetch[(String, String, String)] = Gosub(Gosub(Suspend(Concurrent(List(FetchMany(OneAnd(1,List(2, 3)),ToStringSource$@183bd58)))),<function1>),<function1>)
-
-val result: Future[(String, String, String)] = fetchThree.runA[Future]
-// result: scala.concurrent.Future[(String, String, String)] = List()
+// fetchThree: fetch.Fetch[(String, String, String)] = Gosub(Gosub(Suspend(Concurrent(List(FetchMany(OneAnd(1,List(2, 3)),ToStringSource$@6fb256ea)))),<function1>),<function1>)
 ```
 
-
-When executing the above fetch, note how the three identities get batched and the data source is only queried once. Let's pretend we have a function from `Future[A]` to `A` called `await`.
+When executing the above fetch, note how the three identities get batched and the data source is only queried once.
 
 ```scala
-await(result)
-// [152] Many ToString OneAnd(1,List(2, 3))
-// res4: (String, String, String) = (1,2,3)
+fetchThree.runA[Id]
+// [102] Many ToString OneAnd(1,List(2, 3))
+// res4: cats.Id[(String, String, String)] = (1,2,3)
 ```
 
 ## Parallelism
@@ -161,43 +135,37 @@ If we combine two independent fetches from different data sources, the fetches c
 This time, instead of creating the results with `Query#later` we are going to do it with `Query#async` for emulating an asynchronous data source.
 
 ```scala
-scala> implicit object LengthSource extends DataSource[String, Int]{
-     |   override def fetchOne(id: String): Query[Option[Int]] = {
-     |     Query.async((ok, fail) => {
-     |       println(s"[${Thread.currentThread.getId}] One Length $id")
-     |       ok(Option(id.size))
-     |     })
-     |   }
-     |   override def fetchMany(ids: NonEmptyList[String]): Query[Map[String, Int]] = {
-     |     Query.async((ok, fail) => {
-     |       println(s"[${Thread.currentThread.getId}] Many Length $ids")
-     |       ok(ids.unwrap.map(i => (i, i.size)).toMap)
-     |     })
-     |   }
-     | }
-defined object LengthSource
+implicit object LengthSource extends DataSource[String, Int]{
+  override def fetchOne(id: String): Query[Option[Int]] = {
+    Query.async((ok, fail) => {
+      println(s"[${Thread.currentThread.getId}] One Length $id")
+      ok(Option(id.size))
+    })
+  }
+  override def fetchMany(ids: NonEmptyList[String]): Query[Map[String, Int]] = {
+    Query.async((ok, fail) => {
+      println(s"[${Thread.currentThread.getId}] Many Length $ids")
+      ok(ids.unwrap.map(i => (i, i.size)).toMap)
+    })
+  }
+}
 
-scala> def fetchLength(s: String): Fetch[Int] = Fetch(s)
-fetchLength: (s: String)fetch.Fetch[Int]
+def fetchLength(s: String): Fetch[Int] = Fetch(s)
 ```
 
 And now we can easily receive data from the two sources in a single fetch. 
 
 ```scala
-scala> val fetchMulti: Fetch[(String, Int)] = (fetchString(1) |@| fetchLength("one")).tupled
-fetchMulti: fetch.Fetch[(String, Int)] = Gosub(Gosub(Suspend(Concurrent(List(FetchOne(1,ToStringSource$@183bd58), FetchOne(one,LengthSource$@1d6700b5)))),<function1>),<function1>)
-
-scala> val result = fetchMulti.runA[Future]
-result: scala.concurrent.Future[(String, Int)] = List()
-[163] One ToString 1
+val fetchMulti: Fetch[(String, Int)] = (fetchString(1) |@| fetchLength("one")).tupled
 ```
 
 Note how the two independent data fetches are run in parallel, minimizing the latency cost of querying the two data sources.
 
 ```scala
-scala> await(result)
-[152] One Length one
-res5: (String, Int) = (1,3)
+fetchMulti.runA[Id]
+// [102] One ToString 1
+// [103] One Length one
+// res6: cats.Id[(String, Int)] = (1,3)
 ```
 
 ## Caching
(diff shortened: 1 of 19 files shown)


raulraja:
I looked at the whole set of changes. Excellent work! 
I have two major concerns with this PR.
 
1. We now depend in both places (`Datasource`, `runX`) on `Task`
2. We lost the ability to abstract over the target monad.

I think we need to:

1. Have our own type that wraps Task in this impl to wrap the async behavior we want to support in the `DataSource`
2. Have our own typeclass similar to MonadError but that supports the async behavior for easy conversions to Future, Task, etc... In cases where people would like to target `Id` or similar types where exceptions can't be captured they would be thrown. Cases where the async behavior forces a blocking behavior this would be user provided in the `AsyncMonadError` interface.

I think loosing the abstraction over the target monad it's not acceptable as that is one of the main features for me to use `Fetch`.

I think `Task` is a great type and we should use it internally but not leaking to the user types.
If the user does choose to lift to `Task` explicitly when the computation is evaled then it would be a toll free conversion. so the current behavior you are proposing won't be hurt or damaged in performance in any way.

What I'm trying to say is that there is no need to sacrifice abstracting over the return type while keeping the Task async and concurrent goodness. Thoughts?

dialelo:
Based on your suggestions i reworked the pull request and dropped the coupling with `Task`, making the target monad pluggable again. A summary of the current approach follows, let me know your thoughts.

The `DataSource` typeclass now looks as follows:

```scala
import cats.data.NonEmptyList

trait DataSource[Identity, Result]{
  def fetchOne(id: Identity): Query[Option[Result]]
  def fetchMany(ids: NonEmptyList[Identity]): Query[Map[Identity, Result]]
}
```

The `Query` type represents a query to a remote data source. It supports both synchronous data sources and asynchronous ones, the ADT is roughly as follows:

```scala
sealed trait Query[A] extends Product with Serializable

/** A query that can be satisfied synchronously. **/
final case class Sync[A](action: Eval[A]) extends Query[A]

type Callback[A] = A => Unit
type Errback     = Throwable => Unit

/** A query that can only be satisfied asynchronously. **/
final case class Async[A](action: (Callback[A], Errback) => Unit, timeout: Duration)
    extends Query[A]
```

As you can see, the synchronous queries can be lifted to the target concurrency monad with a plain `.pureEval` and, since we now support asynchronous data sources, a bit more work is needed to lift async queries to an arbitrary monad. Since the asynchronous computation is reduced to the bare minimum (a successful callback and an error callback) it should be trivial to integrate with any [concurrency](https://github.com/47deg/fetch/blob/task-instead-of-eval/shared/src/main/scala/implicits.scala#L28-L36) [monad](https://github.com/47deg/fetch/blob/task-instead-of-eval/monix/shared/src/main/scala/implicits.scala#L28-L37), even ones that [hace synchronous semantics](https://github.com/47deg/fetch/blob/task-instead-of-eval/jvm/src/main/scala/implicits.scala#L30-L51).

I still need to polish things here and there, i will write a task list in the PR description soon.

raulraja:
This is awesome. I don't think we should use await in the docs. I think we should also provide examples for interpreting to Future, Task,  Scalaz Task, etc...  And synchronous ones with big warnings about their blocking and unsafe nature such as Id, Option,  Eval, etc... In any case this is a great advancement in performance and design IMHO

ngbinh:
this PR looks great! 👍

README.md
```diff
@@ -75,88 +81,138 @@ def fetchString(n: Int): Fetch[String] = Fetch(n) // or, more explicitly: Fetch(
...
 
 ```scala
-import fetch.implicits._
 import fetch.syntax._
 
 val fetchOne: Fetch[String] = fetchString(1)
 ```
 
-Now that we have created a fetch, we can run it to a target monad. Note that the target monad (`Eval` in our example) needs to implement `MonadError[M, Throwable]`, we provide an instance for `Eval` in `fetch.implicits._`, that's why we imported it.
+Now that we have created a fetch, we can run it to a `Task`. Note that when we create a task we are not computing any value yet. Having a `Task` instance allows us to try to run it synchronously or asynchronously, choosing a scheduler.
```
tomjadams: I don't see any reference to `Task` in these code snippets. (Not being familiar with the codebase) I'm assuming that `import fetch.implicits._` brings it in (or an implicit conversion from `Fetch` -> `Task`). It might be worth calling that out if so explicitly somewhere in the README.

tomjadams:
Looks good from me, looking forward to playing with it.

dialelo:
Just finished documenting all of this, if you can proofread it and comment will be appreciated!

README.md
```diff
@@ -75,21 +81,29 @@ def fetchString(n: Int): Fetch[String] = Fetch(n) // or, more explicitly: Fetch(
...
 
 ```scala
-import fetch.implicits._
 import fetch.syntax._
 
 val fetchOne: Fetch[String] = fetchString(1)
 ```
 
-Now that we have created a fetch, we can run it to a target monad. Note that the target monad (`Eval` in our example) needs to implement `MonadError[M, Throwable]`, we provide an instance for `Eval` in `fetch.implicits._`, that's why we imported it.
+We'll run our fetches to the ambiend `Id` monad in our examples, let's do some imports.
```
raulraja: Ambient

README.md
```diff
@@ -75,21 +81,29 @@ def fetchString(n: Int): Fetch[String] = Fetch(n) // or, more explicitly: Fetch(
...
-val result: String = fetchOne.runA[Eval].value
-// ToStringSource OneAnd(1,List())
-// result: String = 1
+import cats.Id
+import fetch.unsafe.implicits._
+import fetch.syntax._
 ```
 
-As you can see in the previous example, the `ToStringSource` is queried once to get the value of 1.
+Note that in real-life scenarios you'll want to run a fetch to a concurrency monad, synchronous execution of a fetch
```
raulraja: Would be good to have examples of those Concurrency monads here.

README.md
```diff
@@ -104,41 +118,49 @@ val fetchThree: Fetch[(String, String, String)] = (fetchString(1) |@| fetchStrin
...
 
-And now we can easily recieve data from the two sources in a single fetch. 
+And now we can easily receive data from the two sources in a single fetch. 
 
 ```scala
 val fetchMulti: Fetch[(String, Int)] = (fetchString(1) |@| fetchLength("one")).tupled
 ```
 
-Note how the two independent data fetches are run concurrently, minimizing the latency cost of querying the two data sources. If our target monad was a concurrency monad like `Future`, they'd run in parallel, each in its own logical thread.
+Note how the two independent data fetches are run in parallel, minimizing the latency cost of querying the two data sources.
```
raulraja: Drop 'are'

raulraja:
LGTM. Outstanding job!

README.md
```diff
@@ -11,20 +11,20 @@ A library for Simple & Efficient data access in Scala and Scala.js
 
 Add the following dependency to your project's build file.
 
+For Scala 2.11.x:
+
 ```scala
-"com.fortysevendeg" %% "fetch" %% "0.2.0"
+"com.fortysevendeg" %% "fetch" % "0.2.0"
```
tomjadams: I'm guessing that these versions will change once a new build is made from this PR.

README.md
```diff
@@ -33,8 +33,7 @@ has a latency cost, such as databases or web services.
 
 ## Define your data sources
 
-To tell `Fetch` how to get the data you want, you must implement the `DataSource` typeclass. Data sources have a `fetch` method that
-defines how to fetch such a piece of data.
+To tell Fetch how to get the data you want, you must implement the `DataSource` typeclass. Data sources have `fetchOne` and `fetchMany` methods that define how to fetch such a piece of data.
 
 Data Sources take two type parameters:
```
tomjadams: I would add something into the `Identity` section here like "to fetch the data, this is usually some for of identifier".

README.md
dialelo: Yep, these changes will be released in a 0.3.0 version

README.md
dialelo: Good suggestion, i'll make sure i go into more detail about what I mean by `Identity`.

README.md
```diff
@@ -75,21 +81,29 @@ def fetchString(n: Int): Fetch[String] = Fetch(n) // or, more explicitly: Fetch(
...
-val result: String = fetchOne.runA[Eval].value
-// ToStringSource OneAnd(1,List())
-// result: String = 1
+import cats.Id
+import fetch.unsafe.implicits._
+import fetch.syntax._
 ```
 
-As you can see in the previous example, the `ToStringSource` is queried once to get the value of 1.
+Note that in real-life scenarios you'll want to run a fetch to a concurrency monad such as `Future` or `Task`, synchronous execution of a fetch
```
tomjadams: I think this should move up to underneath the previous paragraph, that way all the code can lice together.

README.md
```diff
@@ -44,24 +43,31 @@ Data Sources take two type parameters:
...
+import cats.data.NonEmptyList
+
 trait DataSource[Identity, Result]{
-  def fetch(ids: NonEmptyList[Identity]): Eval[Map[Identity, Result]]
+  def fetchOne(id: Identity): Query[Option[Result]]
+  def fetchMany(ids: NonEmptyList[Identity]): Query[Map[Identity, Result]]
 }
 ```
 
 We'll implement a dummy data source that can convert integers to strings. For convenience, we define a `fetchString` function that lifts identities (`Int` in our dummy data source) to a `Fetch`.
```
tomjadams: Is it worth showing a non-convenient version of this? i.e. how would it look without the `fetchString` function?

README.md
```diff
@@ -104,41 +118,49 @@ val fetchThree: Fetch[(String, String, String)] = (fetchString(1) |@| fetchStrin
 When executing the above fetch, note how the three identities get batched and the data source is only queried once.
```
tomjadams: Is it fair to say that we could also use `fetchMany` here to achieve the same result? If so, that'd be worth showing, perhaps before the tuples version.

README.md
```diff
@@ -44,24 +43,31 @@ Data Sources take two type parameters:
...
 import cats.std.list._
-
 import fetch._
 
 implicit object ToStringSource extends DataSource[Int, String]{
-  override def fetch(ids: NonEmptyList[Int]): Eval[Map[Int, String]] = {
-    Eval.later({
-      println(s"ToStringSource $ids")
+  override def fetchOne(id: Int): Query[Option[String]] = {
+    Query.later({
```
tomjadams: Probably worth a brief mention of what a `Query` is, and what the operations (`later` and `async` are for. It's implicitly implied  below but never called out, well, not that I've read so far :)

README.md
```diff
@@ -104,41 +118,49 @@ val fetchThree: Fetch[(String, String, String)] = (fetchString(1) |@| fetchStrin
...
+      ok(ids.unwrap.map(i => (i, i.size)).toMap)
     })
   }
 }
 
 def fetchLength(s: String): Fetch[Int] = Fetch(s)
 ```
 
-And now we can easily recieve data from the two sources in a single fetch. 
+And now we can easily receive data from the two sources in a single fetch.
```
tomjadams: Does this have to do with the face that we've marked one as being explicitly async? What if they were both `Query.later`? Or both used `Query.async`? Perhaps instead of "And now..." it could be "Because we've marked one of these data sources as being asynchronous with `Query.later`, Fetch knows that it can execute independent fetches". Or something like that. Could also be reworked into the next paragraph.

README.md
```diff
@@ -104,41 +118,49 @@ val fetchThree: Fetch[(String, String, String)] = (fetchString(1) |@| fetchStrin
...
-// ToStringSource OneAnd(1,List())
-// LengthSource OneAnd(one,List())
-// result: (String, Int) = (1,3)
+fetchMulti.runA[Id]
+// [42] One ToString 1
+// [43] One Length one
+// res7: cats.Id[(String, Int)] = (1,3)
 ```
 
 ## Caching
```
tomjadams: I think it's worth calling out in the opening paragraphs about identity, that you are using the identity and the thing that is fetched for the given the identity interchangeably throughout the examples.

tomjadams:
I've left a whole bunch of comments on the README, mainly from a newbies "I don't understand" POV. Hope they're helpful :)

dialelo:
@tomjadams really valuable feedback, thanks a lot! I'll go through them tomorrow and make sure they get addressed.

tomjadams:
I may have gone a bit nuts :) I just found a few things confusing from a beginners POV.
On Tue, Jun 7, 2016 at 6:24 AM, Aλ <notifications@github.com> wrote:
@tomjadams [https://github.com/tomjadams] really valuable feedback, thanks a lot! I'll go through them tomorrow and make sure they get addressed.

README.md
dialelo: Since identities are batched `fetchMany` is used to get them. The point is that you don't have to manually call `fetchMany`, i do it for you when there is many things of the same type to fetch.

README.md
dialelo: Definitely, i think i'm going to simplify the way of constructing queries. I go into detail about this in the docs but `Query#sync` and `Query#async` should be mentioned in the README.

README.md
dialelo: Done :+1:

README.md
dialelo: is just a matter of using `Fetch#apply` with an identity as shown in the `fetchString` implementation. i want to encourage people to write a little DSL (`getUser`, `getWhatever`) for their fetches since it also helps with inference.

README.md
dialelo: > Does this have to do with the face that we've marked one as being explicitly async?

since one of them is async, we can exploit parallelism even in the `Id` monad. for concurrency monads like Future indepent fetches are run in multiple threads regardless if they are sync or async. 

i'm not sure introducing `Query#async` here is a good idea, maybe i should rework to use `Query#sync` and see how `Id` doesn't parallelize them; then run to `Future` and see the parallelism.

> What if they were both Query.later? 

they'll run sequentially in `Id`, potentially in parallel in a concurrency monad like `Future`.

> Or both used Query.async

in parallel in `Id`, potentially in parallel in `Future` et al

README.md
dialelo: good catch, i'll reword that since it'll probably cause confusion.

dialelo:
I've clarified the `Query` API and its documentation. I'll leave testing monix more thoroughly to #32, since I plan to run the same property-based tests across all target types. @raulraja if you give me the thumbs up I'll merge this.

raulraja:
LGTM!

raulraja:
Again, amazing job!

dialelo:
:shipit:

PR #57, opened 2016-06-06
Add code coverage reporting

dialelo:
@rafaparadela plz take a look.


dialelo:
Oops, looks like is not compatible with Scala.js.

dialelo:
@rafaparadela fixed the problems with Scala.js, plz take a look!

rafaparadela:
Nice! LGTM

PR #66, opened 2016-07-26
Improved reporting and error handling

dialelo:
@raulraja please take a look and let me know your thoughts.
- [x] Ensure that `fetchMany` is never called with duplicated identities

I've modified the interpreter for storing information about execution rounds: the request, result and timing of each request is stored in the environment. Cached requests do no longer count as rounds, so it maps better to the mental model of a fetch execution.

I also modified the error type to be a `FetchError` (which is now open to extension) and discerned between a single identity that's missing `NotFound` and potentially multiple missing identities by batching or parallel fetching `MissingIdentities`. I also wrapped exceptions thrown in data source client code in a `FetchException` type.

This will require a few changes in the Fetch exercises, I can do that after merging this PR.

### DIFF
--- a/shared/src/main/scala/fetch.scala
+++ b/shared/src/main/scala/fetch.scala
@@ -70,6 +70,11 @@ sealed trait FetchQuery[I, A] extends FetchRequest {
   def identities: NonEmptyList[I]
 }
 
+trait FetchError extends Throwable with Product with Serializable
+case class NotFound(env: Env, request: FetchOne[_, _])                          extends FetchError
+case class MissingIdentities(env: Env, missing: Map[DataSourceName, List[Any]]) extends FetchError
+case class FetchException(err: Throwable)                                       extends FetchError
+
 /**
   * Primitive operations in the Fetch Free monad.
   */
@@ -109,7 +114,7 @@ final case class Concurrent(as: List[FetchQuery[_, _]])
     as.forall(_.fullfilledBy(cache))
   }
 }
-final case class FetchError[A](err: Throwable) extends FetchOp[A]
+final case class Thrown[A](err: Throwable) extends FetchOp[A]
 
 object `package` {
   type DataSourceName     = String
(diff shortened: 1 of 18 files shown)


shared/src/main/scala/fetch.scala
```diff
@@ -60,22 +60,30 @@ object Query {
...
   def fullfilledBy(cache: DataSourceCache): Boolean
+}
+
+sealed trait FetchQuery[I, A] extends FetchRequest {
   def missingIdentities(cache: DataSourceCache): List[I]
   def dataSource: DataSource[I, A]
   def identities: NonEmptyList[I]
 }
 
+trait FetchError extends Throwable with Product with Serializable
```
raulraja: I'm not sure FetchError is a good name, I'd call it FetchException as Error indicates potentially a real JVM Error when I read it and I would not dare to catch that.

shared/src/main/scala/interpreters.scala
```diff
@@ -54,19 +49,83 @@ trait FetchInterpreters {
...
+                val startRound = System.nanoTime()
+                val cache      = env.cache
+                val newIds     = many.missingIdentities(cache)
+                val result     = ids.unwrap.flatMap(id => cache.get(ds.identity(id)))
+
+                if (newIds.isEmpty)
+                  M.pure(env -> result)
+                else {
+                  M.flatMap(M.runQuery(ds
+                            .asInstanceOf[DataSource[I, A]]
```
raulraja: Is there a way to get rid of these casts? If we are not sure we should add an issue to get it cleaned up a bit.

raulraja:
Sorry for the late review, so far what I've seen LGTM, Can't wait to get the Fetch lib published in Scala Exercises :)

shared/src/main/scala/interpreters.scala
dialelo: Opened #67 to address that.

PR #70, opened 2016-11-02
Finish cats 0.7.2 update and change `deps` method

peterneyens:
I replaced the `Cont[FM, ?]` monad trick with an interpreter to `EitherT[Writer[List[FetchOp[_]], ?]. (), ?]`.


dialelo:
Awesome work :clap: :clap:

PR #73, opened 2016-11-04
Refactor and use some more cats features

peterneyens:
- Refactor interpreter: split out handling of `FetchOne`, `FetchMany`
  and `Concurrent`, use `XorT` in `processMany` to reduce nested folds,
  use `Validated` in `processMany` and `processConcurrent` to get
  missing ids/identities or some result.
- (Prematurely ?) optimize `Fetch.join`
- Replaced custom implementations of `map2`, `sequence` and `traverse`
  by using `Applicative[Fetch]`
- ...

I moved some methods around in the `Fetch` object and with the
extraction of the three methods in the interpreter, so the diff looks
much bigger than it actually is.


dialelo:
Amazing work :clap: :clap:, learning a lot form these contributions.

PR #76, opened 2016-11-08
Support a maximum batch size per `DataSource`

peterneyens:
- Add the `maxBatchSizePhase` interpreter which splits fetches to data sources with a maximum batch size in to multiple sequential fetches. This required a change to `Concurrent` to return a `InMemoryCache` instead of a `DataSourceCache` making it possible to combine multiple caches for sequential batches.
- Refactor the `processConcurrent` method which is used in the `coreInterpreter` (previously called `interpreter`) by using `XorT` instead of nested `if` / `else` and `Option.fold` and by extracting some helper methods.
- Give some members of the `Fetch` constructors a better name (eg. `FetchMany.as` to `FetchMany.ids`)
- Create the `Fetch.multiple` method which makes it possible to create a `Fetch[List[...]]` directly (using `FetchMany` underneath). This can currently (and previously) be done by using `sequence` or `traverse`, but `multiple` doesn't have the overhead of inspecting the Fetch structure to find independent fetches which can be parallelized.

### DIFF
--- a/shared/src/main/scala/cache.scala
+++ b/shared/src/main/scala/cache.scala
@@ -29,6 +29,8 @@ trait DataSourceCache {
       case (acc, (i, a)) => acc.update(ds.identity(i), a)
     })
   }
+
+  def contains(k: DataSourceIdentity): Boolean = get(k).isDefined
 }
 
 /**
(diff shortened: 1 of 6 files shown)


dialelo:
Outstanding work, well done! :shipit:

Issue #68, opened 2016-09-02
Update to latest cats and monix


ngbinh:
Now that we have cats `0.7.2` and monix final `2.0.0` released. Please consider upgrade the dependencies and publish a stable version. I tried it myself but got stuck on all the `tailRecM` stuff on the new cats Monad.

dialelo:
Sorry about not releasing, have been very busy and haven't had time to devote to open source. Will look into updating dependencies and releasing as soon as i have time.

ngbinh:
No worries. It's great to have opensource libraries like fetch. I would
love to help if I can.

dialelo:
It'll require some work since the latest cats changes make it more difficult to move computations around using `Eval`.

dialelo:
Thanks for the kind words!

RomanIakovlev:
Hey, it would be super-awesome to get the Fetch with Monix 2.x. It's a total deal-breaker at the moment. It seems to be in the work, but if any help is needed, please do let know.

dialelo:
It has already been merged, will release a new version early next week.

juanpedromoreno:
Fetch 0.4.0 is out!

https://github.com/47deg/fetch/releases/tag/v0.4.0

Thanks @peterneyens !

PR #80, opened 2016-11-09
Update cats to 0.8.1 and add Scala 2.12 support

### LINKED ISSUE #72: Publish version for Scala 2.12

peterneyens:
I updated cats to 0.8.1 (and monix to the compatible 2.1.0).
I think that all our dependencies and plugins now make it possible to also release for Scala 2.12.0.

The cats changes are mostly changing `Xor`/`XorT` to `Either`/`EitherT` and removing the `RecursiveTailRecM` evidence. I added the `FromMonadError` class which makes it easier to implement `FetchMonadError[M]` given a `MonadError[M, Throwable]` (mostly so that we are spared from reimplementing `tailRecM`). 

I updated the scalafmt version which triggered some code reformatting. There is apparently an issue with sbt and scalafmt (https://github.com/sbt/sbt/issues/2786) which keeps us from using sbt 0.13.13.

Fixes #72 and #68.

@dialelo and @juanpedromoreno, would you like to take a look?


build.sbt
```diff
@@ -28,10 +29,11 @@ lazy val buildSettings = Seq(
 lazy val commonSettings = Seq(
   resolvers += Resolver.sonatypeRepo("releases"),
   libraryDependencies ++= Seq(
-    "org.typelevel" %%% "cats" % "0.7.2",
+    "org.typelevel" %%% "cats-core" % "0.8.1",
+    "org.typelevel" %%% "cats-free" % "0.8.1",
```
peterneyens: I need to add cats-kernel and cats-macros here as well.

dialelo:
LGTM :shipit:

PR #82, opened 2016-11-09
Update cats version to 0.8.1 and monix version to 2.1.0

ilya-murzinov:


peterneyens:
Thanks @ilya-murzinov, but I am updating cats and monix (and some other stuff) in #80.

ilya-murzinov:
OK :)

PR #83, opened 2016-11-10
Aggregates fetch-monix in the root module

juanpedromoreno:
This PR includes `fetch-monix` module as a part of the root module, publishing it with the rest of the modules.

Please, @dialelo review at your convenience. Thanks.


dialelo:
Good catch, thanks!

PR #84, opened 2016-11-14
New implementation parallel joins

peterneyens:
This changes the way fetches are inspected to parallelize independent fetches. It does this by adding a new `Join` construct to the `Fetch` structure. This `Join` is parallelized in a new intermediate interpreter which inspects the actual `Free` constructs and replaces cached fetches by `Free.Pure`.

It does play a bit fast and loose with the laws with this fiddling with `Free` under the hood, but it now executes code in a `fetch.map(...)` or a `fetch.flatMap(...)` only once. 

(The other PR updating cats and adding Scala 2.12 support should be merged first).


peterneyens:
I'll rebase this on the current master.

dialelo:
LGTM :shipit:

PR #86, opened 2016-11-14
Fixes Cross Publishing

juanpedromoreno:
I've just realized we should cross-publish with `+publishSigned` instead of just `publishSigned`, now that we have cross compatibility through scala `2.11` and `2.12`. Also, if we don't apply `allSettings` to the root project, the cross versioning starts with scala `2.10`.

This PR comes to fix it. 

@dialelo  @peterneyens  could you take a look? Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,11 +10,11 @@ before_install:
    -in secring.gpg.enc -out secring.gpg -d; fi
 script:
 - sbt coverage 'fetchJVM/test' 'fetchJVM/coverageReport'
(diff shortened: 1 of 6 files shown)


dialelo:
Good catch! LGTM

peterneyens:
It seems I didn't check the documentation for 2.11 after the cats update, I think we are missing `import cats.syntax.either._` in _docs.md_.

peterneyens:
:+1: Thanks for cleaning up after me :blush:

PR #88, opened 2016-11-15
Cross-compile to Scala 2.10.x

adelbertc:
Only source change needed was value classes need a public
val member.

On running tests locally it seems to have reformatted a bunch of
stuff. In the interest of not confusing other contributors I've
left reformats as is.


dialelo:
Thanks!

dialelo:
The next 0.4.1-SNAPSHOT will be published for Scala 2.10 by Travis automatically, thanks @adelbertc!

PR #89, opened 2016-11-21
Make `DataSource#name` mandatory

### LINKED ISSUE #77: Enforce users to declare a name for every DataSource
Since it turns out that there is no way to get a unique name with the current class name approach, we should require `DataSource` implementations to define the `name` method from now on.

dialelo:
Closes #77


dialelo:
@juanpedromoreno @peterneyens @raulraja take a look, we may want to bump the version number since I guess this is binary incompatible with previous 0.4.x releases.

PR #90, opened 2016-11-21
Fixes Compatibility with Scala 2.10

juanpedromoreno:
This PR fixes/adds:

* Adds scala 2.10 to the travis file to run the tests for this scala version.
* Fixes compatibility with Scala 2.10 adding the `paradise` compiler plugin (needed when using with the `kind-projector` and Scala 2.10).
* Fixes docs (divergence implicits issue) for Scala 2.10 with `-Xdivergence211` compiler flag.
* Upgrades `sbt-catalysts-extras`, squeezing all the plugin dependencies into a single one.
* Cats, kind-projector, monix and scalatest dependencies are using the same versions as before, but in this case, through the `sbt-catalysts-extras` plugin.

Please, @dialelo  @raulraja  could you review? Thanks!


raulraja:
LGTM! /cc @adelbertchang

adelbertc:
Hurrah!

PR #95, opened 2016-12-02
Reduce stack consumption

dialelo:
- [x] Query aggregation
- [x] Join node simplification

Tried some extreme cases and found a couple of parts in the code where we could reduce stack consumption. Unfortunately the tests are failing due to a timeout in Travis CI, looking into it right now.


raulraja:
This looks great, looks like codecov says it decreased in 3% though and bumps it below 80% which we may want to address if it's something we can do as part of this PR. This improvements deserves a new release :clap:

shared/src/test/scala/FetchBatchingTests.scala
```diff
@@ -0,0 +1,145 @@
...
+
+  "Very deep fetches don't overflow stack or heap" in {
+    import cats.syntax.traverse._
+
+    val depth = 200
+    val fetch: Fetch[List[Int]] = (1 to depth).toList
+      .map((x) => (0 until 2).toList.traverse(fetchBatchedDataSeq))
+      .foldLeft(
+        Fetch.pure(List.empty[Int])
+      )((acc, f) => acc.flatMap((x) => f))
```
peterneyens: This `foldLeft` with `pure` and `flatMap` could also be written using the monadic fold (`Foldable#foldM`).

Something like (might need some type annotations): 
```scala
import cats.instances.list._
import cats.syntax.foldable._

.foldM(List.empty[Int])((_, f) => f)
// or using foldMapM (this does combine all the lists though)
.foldMapM(identity)
```

shared/src/test/scala/FetchTests.scala
```diff
@@ -98,916 +98,3 @@ object TestHelper {
...
-  "`runA` syntax is equivalent to `Fetch#run`" in {
-
-    val rf1 = Fetch.run[Future](1.fetch)
-    val rf2 = 1.fetch.runA[Future]
-
-    ME.map2(rf1, rf2)(_ shouldEqual _)
-  }
-}
-
-class FetchTests extends AsyncFreeSpec with Matchers {
```
peterneyens: Have we dropped these ?
Seems we are removing around 550 lines in this PR, which is more or less the same as these tests ?

dialelo:
@peterneyens i added the lost tests back and changed the `foldLeft` usage for `foldM`, check it out please!

dialelo:
:shipit:

PR #96, opened 2016-12-02
Add example subproject with doobie and http4s examples

peterneyens:
Add a subproject with examples using doobie and http4s.


dialelo:
Awesome, love it! Really helps having example projects to complement the documentation.

PR #99, opened 2017-01-26
Update Cats and Monix

### LINKED ISSUE #98: Update to cats 0.9.0
would be nice because it is currently blocking us.

peterneyens:
Some code clean up using `Reducible#reduceMapM` and the now covariant `NonEmptyList`.

The http4s-circe version still depends on circe 0.6.1 and thus cats 0.8.1. It is only used in the example project of Fetch, so it doesn't matter here, but we should probably update sbt-catalysts-extras once a new http4s version comes out.

Resolves #98


dialelo:
Awesome, LGTM :shipit:

PR #105, opened 2017-02-21
Set version to 0.5.0 for publishing under new group id

dialelo:
Since the org was renamed after 0.5.0 was published, we need to release the 0.5.0 artifacts with the new organization name. Merging this PR in master will publish them automatically.

Closes #103


juanpedromoreno:
`fetch` has been already published under `com.47deg` group id ;)

PR #109, opened 2017-03-09
Add curried apply to Fetch object

AlecZorab:
very minor convenience, allows users to do

```scala
def getUser(id: UserId) = Fetch[User](id)
```


dialelo:
Thanks!

PR #110, opened 2017-03-09
add batchingOnly convenience method

AlecZorab:
- similarly to batchingNotSupported, but for sources that only do
batching (or for which batching is just as natural)


dialelo:
Nice, thanks for taking the time to contribute!

PR #113, opened 2017-03-14
Allow sequential and parallel batches

peterneyens:
I made it possible to configure if you want to execute batches to a data source sequentially or in parallel.

By default they will be parallel, so we should mention in the next release notes that people relying on the sequential execution of the batches should update their data sources to specify this sequential execution.


dialelo:
👏 👏 👏 Amazing work, well done!

PR #114, opened 2017-03-17
0.6.0 release

dialelo:
Title says it all, had to revert #109 due to clashes with the `Fetch` companion's apply method. Other than that is just documentation for the new features and a changelog update.

### DIFF
--- a/docs/src/main/tut/docs.md
+++ b/docs/src/main/tut/docs.md
@@ -29,13 +29,13 @@ we read) concerns.
 To begin, add the following dependency to your SBT build file:
 
 ```scala
-"com.47deg" %% "fetch" % "0.5.0"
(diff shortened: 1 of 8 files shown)


docs/src/main/tut/docs.md
```diff
@@ -582,6 +597,83 @@ val fetchSameTwice: Fetch[(User, User)] = for {
 fetchSameTwice.runA[Id](ForgetfulCache())
 ```
 
+# Batching
+
+As we have learned, Fetch performs batched requests whenever it can. It also exposes a couple knobs
+for tweaking the maximum batch size and whether multiple batches are run on parallel or sequentially.
```
peterneyens: `on` -> `in` ?

dialelo: Good catch!

PR #117, opened 2017-04-19
Removes dup doc files

juanpedromoreno:
This PR removes some unneeded files and it also removes the `README.md` file inside the tut directory. Hence, tut will compile the file in the root folder instead.

Please, @fedefernandez could you review? Thanks!


project/ProjectPlugin.scala
```diff
@@ -52,8 +52,8 @@ object ProjectPlugin extends AutoPlugin {
       aggregate in doc := true)
 
     lazy val readmeSettings: Seq[Def.Setting[_]] = tutSettings ++ commonTutSettings ++ Seq(
-      tutSourceDirectory := baseDirectory.value,
-      tutTargetDirectory := baseDirectory.value.getParentFile,
+      tutSourceDirectory := baseDirectory.value.getParentFile,
```
fedefernandez: Maybe we could use here `(baseDirectory in LocalRootProject).value` instead

project/ProjectPlugin.scala
```diff
@@ -52,8 +52,8 @@ object ProjectPlugin extends AutoPlugin {
...
+      tutTargetDirectory := baseDirectory.value.getParentFile / "target",
```
fedefernandez: Same as above

project/ProjectPlugin.scala
```diff
@@ -82,6 +82,7 @@ object ProjectPlugin extends AutoPlugin {
         ScalaJSBadge.apply(_),
         GitHubIssuesBadge.apply(_)
       ),
+      orgSupportedScalaJSVersion := Some("0.6.15"),
```
fedefernandez: :+1:

PR #127, opened 2017-07-18
Add timeout capability to Future implementation of FetchMonadError

justinhj:
The attached PR adds timeout capability to FetchMonadError[Future]

I also modified Query.Sync to return a completed future which I think is the intent rather than run it in a Future

All tests pass but I didn't yet create tests for the timeout, I can do so if the PR looks like something you might want

Only thing I'm not sure about is whether the use of synchronize is needed. I'm not sure but I think Promise may be safe to use from multiple threads because it uses an AtomicReference but didn't find any confirmation of that.

### DIFF
--- a/jvm/src/test/scala/FutureTimeoutTests.scala
+++ b/jvm/src/test/scala/FutureTimeoutTests.scala
@@ -20,7 +20,6 @@ import scala.concurrent._
 import scala.concurrent.duration._
 import org.scalatest._
 import cats.data.NonEmptyList
-import fetch._
 import fetch.implicits._
 
 // Note that this test cannot run on Scala.js
@@ -35,9 +34,7 @@ class FutureTimeoutTests
   implicit override def executionContext: ExecutionContext = ExecutionContext.Implicits.global
 
   case class ArticleId(id: Int)
-  case class Article(id: Int, content: String) {
-    def author: Int = id + 1
-  }
+  case class Article(id: Int, content: String)
 
   def article(id: Int)(implicit DS: DataSource[ArticleId, Article]): Fetch[Article] =
     Fetch(ArticleId(id))
@@ -56,7 +53,7 @@ class FutureTimeoutTests
       batchingNotSupported(ids)
   }
 
-  "fetchfuture" should "fail with timeout when a datasource does not complete in time" in {
+  "FetchMonadError[Future]" should "fail with timeout when a datasource does not complete in time" in {
 
     implicit val dsWillTimeout = ConfigurableTimeoutDatasource(250 milliseconds, 750 milliseconds)
 
--- a/jvm/src/test/scala/FutureTimeoutTests.scala
+++ b/jvm/src/test/scala/FutureTimeoutTests.scala
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+package fetch
+
 import scala.concurrent._
 import scala.concurrent.duration._
 import org.scalatest._
@@ -23,9 +25,14 @@ import fetch.implicits._
 
 // Note that this test cannot run on Scala.js
 
-class FutureTimeoutTests extends AsyncFlatSpec with Matchers with OptionValues with Inside with Inspectors {
+class FutureTimeoutTests
+    extends AsyncFlatSpec
+    with Matchers
+    with OptionValues
+    with Inside
+    with Inspectors {
 
-  implicit override def executionContext : ExecutionContext = ExecutionContext.Implicits.global
+  implicit override def executionContext: ExecutionContext = ExecutionContext.Implicits.global
 
   case class ArticleId(id: Int)
   case class Article(id: Int, content: String) {
@@ -37,7 +44,8 @@ class FutureTimeoutTests extends AsyncFlatSpec with Matchers with OptionValues w
 
   // A sample datasource with configurable delay and timeout
 
-  case class ConfigurableTimeoutDatasource(timeout: Duration, delay: Duration) extends DataSource[ArticleId, Article] {
+  case class ConfigurableTimeoutDatasource(timeout: Duration, delay: Duration)
+      extends DataSource[ArticleId, Article] {
     override def name = "ArticleFuture"
     override def fetchOne(id: ArticleId): Query[Option[Article]] =
       Query.async((ok, fail) => {
@@ -53,7 +61,7 @@ class FutureTimeoutTests extends AsyncFlatSpec with Matchers with OptionValues w
     implicit val dsWillTimeout = ConfigurableTimeoutDatasource(250 milliseconds, 750 milliseconds)
 
     val fetch: Fetch[Article] = article(1)
-    val fut: Future[Article] = Fetch.run[Future](fetch)
+    val fut: Future[Article]  = Fetch.run[Future](fetch)
 
     recoverToSucceededIf[TimeoutException] {
       fut
@@ -66,9 +74,9 @@ class FutureTimeoutTests extends AsyncFlatSpec with Matchers with OptionValues w
     implicit val dsWillTimeout = ConfigurableTimeoutDatasource(750 milliseconds, 250 milliseconds)
 
     val fetch: Fetch[Article] = article(1)
-    val fut: Future[Article] = Fetch.run[Future](fetch)
+    val fut: Future[Article]  = Fetch.run[Future](fetch)
 
-    fut.map{_ shouldEqual Article(1, "An article with id 1")}
+    fut.map { _ shouldEqual Article(1, "An article with id 1") }
   }
 
   it should "not fail with timeout when infinite timeout specified" in {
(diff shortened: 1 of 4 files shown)


justinhj:
Removed synchronization on Promise and simplified the PR

justinhj:
Have added tests for the two cases of a future failing because it timed out and succeeding because it finished before the timeout occurred.

shared/src/test/scala/FutureTimeoutTests.scala
```diff
@@ -0,0 +1,87 @@
...
+    def author: Int = id + 1
+  }
+
+  // A sample datasource with configurable delay and timeout
+
+  case class ConfigurableTimeoutDatasource(timeout: Duration, delay: Duration) extends DataSource[ArticleId, Article] {
+    override def name = "ArticleFuture"
+    override def fetchOne(id: ArticleId): Query[Option[Article]] =
+      Query.async((ok, fail) => {
+        Thread.sleep(delay.toMillis)
```
peterneyens: `Thread.sleep` doesn't work on ScalaJS, which is the reason that Travis is complaining.

shared/src/main/scala/implicits.scala
```diff
@@ -16,29 +16,71 @@
...
+  val timer: Timer = new Timer()
+
+  implicit def fetchFutureFetchMonadError(implicit ec: ExecutionContext): FetchMonadError[Future] =
     new FetchMonadError.FromMonadError[Future] {
       override def runQuery[A](j: Query[A]): Future[A] = j match {
-        case Sync(e) => Future(e.value)
-        case Async(ac, timeout) => {
+
+        case Sync(e) =>
+          Future.successful({e.value})
```
peterneyens: Evaluating `e` could potentially throw an exception and those aren't caught by `Future.successful`.

shared/src/test/scala/FutureTimeoutTests.scala
```diff
@@ -0,0 +1,87 @@
...
+      }, timeout)
+    override def fetchMany(ids: NonEmptyList[ArticleId]): Query[Map[ArticleId, Article]] =
+      batchingNotSupported(ids)
+  }
+
+  "fetchfuture" should "fail with timeout when a datasource does not complete in time" in {
+
+    implicit val dsWillTimeout = ConfigurableTimeoutDatasource(250 milliseconds, 750 milliseconds)
+
+    def article(id: Int): Fetch[Article] = Fetch(ArticleId(id))
```
peterneyens: If you added an implicit `DataSource[ArticleId, Article]` you wouldn't need to implement this method in all the tests.

```scala
def article(id: Int)(implicit DS: DataSource[ArticeId, Article]): Fetch[Article] =
  Fetch(ArticleId(id))  // or Fetch(ArticleId(id))(DS)
```

shared/src/test/scala/FutureTimeoutTests.scala
```diff
@@ -0,0 +1,87 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import scala.concurrent._
+import scala.concurrent.duration._
+import org.scalatest._
+import cats.data.NonEmptyList
+import scala.language.postfixOps
```
peterneyens: You shouldn't need this as we have `"-language:postfixOps"` as a `scalacOption`.

shared/src/test/scala/FutureTimeoutTests.scala
```diff
@@ -0,0 +1,91 @@
...
+    Fetch(ArticleId(id))
+
+  // Simulate Thread.sleep for the purposes of making this test work in Scala.js
+
+  def sleep(ms: Long): Unit = {
+    println("start sleep")
+    val expTime = System.nanoTime + ms*1000000
+    while(System.nanoTime < expTime) {}
+    println("sleep finished")
+  }
```
peterneyens: I am not sure if we can actually test the time out in ScalaJS.

I think that if you block like this, you block everything in JS, so the `TimerTask` cannot finish the promise with a `TimeoutExecption` (which is happening in the failing test on Travis).

We could revert back to using `Thread.sleep` and add this as a JVM only test, by moving it to `jvm/src/test/scala`, what do you think?

justinhj: Yeah I think that makes sense. I've been trying to think of a way to isolate the timeout behaviour in a way that can be demonstrated to work with scala js but haven't come up with a way yet.

justinhj:
I've updated the PR so that the tests that require non-blocking sleep are now in the JVM module. Seems travis build is failing but this is due to a timeout building the docs

shared/src/main/scala/implicits.scala
```diff
@@ -16,29 +16,65 @@
...
+                def run() : Unit = {
+                  p.tryFailure(new TimeoutException())
+                }
+              }
+
+              // Start the timeout Timer
+              timer.schedule(timerTask, timeout.toMillis)
+
+              // Execute the user's action
+              ec.execute(() => {
```
peterneyens: This conversion (and the one below) to `Runnable` only works in 2.12 I think which added support for Java 8 SAMs.

justinhj: Oops my bad. IntelliJ suggested it and it seemed like a good idea. I'll change it back

justinhj:
I think everything looks good now, just getting a timeout with Travis building the docs

peterneyens:
There definitely is something strange going on when combining `Timer` and tut.

Commenting out `implicits.timer` (and as a consequence `timer.schedule(...)`) makes tut run fine.

A minimal tut file which makes tut hang (notice that we aren't even using the time out here, since we are using `Query.sync`):

```scala
// ```tut
import scala.concurrent.{Await, Future}
import scala.concurrent.duration.Duration
import scala.concurrent.ExecutionContext.Implicits.global

import fetch.{FetchMonadError, Query}
import fetch.implicits._

val query: Query[Option[Int]] = Query.sync { Some(1) }

val future: Future[Option[Int]] = FetchMonadError[Future].runQuery(query)
Await.result(future, Duration.Inf)
// ```
```

If I change the `Sync` from `Future(...)` to the line below, tut runs fine for the minimal file, not for the complete _docs.md_ file though ...
```scala
case Sync(e) =>
  cats.ApplicativeError[Future, Throwable].catchNonFatalEval(e)
```

It seems like I need to do some more investigating to see what is going on ...

justinhj:
Ah timer is supposed to run on a daemon thread and if it doesn't then it can stop a program from terminating. I'll check the code and see if I'm doing that correctly

justinhj:
I've updated so Timer is on a daemon thread and also created lazily so it is not initialized by programs that don't use it

jvm/src/test/scala/FutureTimeoutTests.scala
```diff
@@ -0,0 +1,92 @@
...
+    with Matchers
+    with OptionValues
+    with Inside
+    with Inspectors {
+
+  implicit override def executionContext: ExecutionContext = ExecutionContext.Implicits.global
+
+  case class ArticleId(id: Int)
+  case class Article(id: Int, content: String) {
+    def author: Int = id + 1
```
peterneyens: I seems this `author` method isn't used, so we can drop it, right?

shared/src/main/scala/implicits.scala
```diff
@@ -16,29 +16,69 @@
...
+                def run() : Unit = {
+                  ac(p.trySuccess, p.tryFailure)
+                }
+              })
+
+            // No timeout 
+            case _ =>
+
+              // Execute the user's action
+              ec.execute(new Runnable {
```
peterneyens: Nitpicking, we can create this `Runnable` once before the `timeout match { ... }`.
Even more nitpicking, we can lose the curly braces around `ac(p.trySuccess, p.tryFailure)`.

shared/src/main/scala/implicits.scala
```diff
@@ -16,29 +16,69 @@
...
 import cats.instances.future._
 import scala.concurrent.{ExecutionContext, Future, Promise}
 
 object implicits {
 
-  implicit def fetchFutureFetchMonadError(
-      implicit ec: ExecutionContext
-  ): FetchMonadError[Future] =
+  // Shared Timer object to schedule timeouts
+  private[fetch] lazy val timer: Timer = new Timer(true)
```
peterneyens: We should probably name the daimon thread as well, something like "fetch-future-timeout-daemon" ?

justinhj: yeah good idea, will help avoid confusion when people look at running threads and don't know what it's for

shared/src/main/scala/implicits.scala
justinhj: Done, maybe there's a better naming for the runnable than 'runnable', suggestions welcome

PR #137, opened 2017-10-06
Update dependencies (Cats 1.0.0-MF)

peterneyens:
This PR upgrades Fetch to Cats 1.0.0-MF (and monix to 3.0.0-M1, catbird to 0.18.0).
We are now also using sbt 1.0.2 (thanks to @suhasgaddam for his help here and in sbt-org-policies).

### DIFF
--- a/docs/src/main/tut/docs.md
+++ b/docs/src/main/tut/docs.md
@@ -133,16 +133,11 @@ val userDatabase: Map[UserId, User] = Map(
 implicit object UserSource extends DataSource[UserId, User]{
   override def name = "User"
 
-  override def fetchOne(id: UserId): Query[Option[User]] = {
-    Query.sync({
-	  latency(userDatabase.get(id), s"One User $id")
-    })
-  }
-  override def fetchMany(ids: NonEmptyList[UserId]): Query[Map[UserId, User]] = {
-    Query.sync({
-	  latency(userDatabase.filterKeys(ids.toList.contains), s"Many Users $ids")
-    })
-  }
+  override def fetchOne(id: UserId): Query[Option[User]] =
+    Query.sync(latency(userDatabase.get(id), s"One User $id"))
+
+  override def fetchMany(ids: NonEmptyList[UserId]): Query[Map[UserId, User]] =
+    Query.sync(latency(userDatabase.filterKeys(ids.toList.toSet), s"Many Users $ids"))
 }
 ```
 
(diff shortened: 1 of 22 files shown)


docs/src/main/tut/docs.md
```diff
@@ -1184,8 +1161,6 @@ Note that Cats' typeclass hierarchy is expressed with inheritance and methods fr
 We make use of the `FromMonadError` class below, making it easer to implement `FetchMonadError[Task]` given a `MonadError[Task, Throwable]` which we can get from the _monix-cats_ projects.
 
 ```tut:silent
-import monix.cats._
-
 implicit val taskFetchMonadError: FetchMonadError[Task] =
   new FetchMonadError.FromMonadError[Task] {
     override def runQuery[A](q: Query[A]): Task[A] = queryToTask[A](q)
```
fedefernandez: Has this method been removed [here](https://github.com/47deg/fetch/pull/137/files?diff=split#diff-6d6feb92c529f59043fb943fd182a5e2L30)? Should this be updated?

shared/src/main/scala/fetch.scala
```diff
@@ -205,9 +162,9 @@ object `package` {
      * results. It implies concurrent execution of fetches.
      */
     def join[A, B](fl: Fetch[A], fr: Fetch[B]): Fetch[(A, B)] =
-      Free.liftF(Join(fl, fr))
+      Free.liftF[FetchOp, (A, B)](Join(fl, fr))
 
-    class FetchRunner[M[_]] {
+    private[fetch] class FetchRunner[M[_]](private val dummy: Boolean = true) extends AnyVal {
```
fedefernandez: I'm curious, why we need this `dummy` value?

peterneyens: That is a trick to make `FetchRunner` a value class that is used in Cats as well.

docs/src/main/tut/docs.md
peterneyens: I removed the method from `FromMonadError` because it is inherited from `FetchMonadError`.

shared/src/main/scala/fetch.scala
fedefernandez: Great, thanks for the responses

peterneyens:
II think if we merge this, merge master in #135 it will pass as well.
After that I'll open an new PR for a 0.7.0 release.

PR #138, opened 2017-10-06
update name of twitter module

chris-vale-ck:
rename module to `fetch-twitter`


peterneyens:
Thanks @chris-vale-ck!

Issue #148, opened 2018-03-04
Turn `DataSourceCache` type class parametric to F[_]


fedefernandez:
The current implementation assumes that a Cache will return its results blocking if it needs to go async to fetch them and that is not good.

purrgrammer:
In the implementation I'm working on `DataSource` is parameterized to `F[_] : Effect` in both `DataSource#fetchOne` and `DataSource#fetchMany`. As you suggest the `DataSourceCache` should be parameterized to `F[_] : Effect` too, I'll make sure to rework it.

PR #155, opened 2018-08-31
Redesign

purrgrammer:
This is still a WIP but I managed to rewrite the core `Fetch` execution model using an approach much closer to [the original paper](https://simonmar.github.io/bib/papers/haxl-icfp14.pdf). The goal was to modernize the library and use the primitives from `cats-effect`.

I removed the `Query` type and I'm instead using `ConcurrentEffect` in data sources, it looks like:

```scala
import cats.temp.par._
import cats.effect._

trait DataSource[I, A]{
  def name: String

  def fetch[F[_] : ConcurrentEffect : Par](id: I): F[Option[A]]

  /* automatically implemented in terms of `fetch` and parTraverse, user can override */
  def batch[F[_] : ConcurrentEffect : Par](ids: NonEmptyList[I]): F[Map[I, A]] 
}
```

`Par` is a typeclass that eases using `cats.Parallel`, taking only one type parameter instead of two. It comes from [cats-par](https://github.com/ChristopherDavenport/cats-par).

I've gotten rid of `Free` in our code, since it didn't allow us to explore both sides of a computation in the way this encoding does. Now a description of a `Fetch` is encoded in the `FetchResult` type which can be `Done` or `Blocked` waiting for a computation to finish before continuing. Having our own representation of the primitives we use from `Free` (pure, flatMap) allows us to explore (and combine) both sides of a computation and group their requests.

`DataSourceCache` is also implemented in terms of `ConcurrentEffect` now:

```scala
trait DataSourceCache {
  def lookup[F[_] : ConcurrentEffect, I, A](i: I, ds: DataSource[I, A]): F[Option[A]]
  def insert[F[_] : ConcurrentEffect, I, A](i: I, v: A, ds: DataSource[I, A]): F[DataSourceCache]
}
```

which should solve #148 and allow users to use caches that perform `IO` operations. I still need to write a proof-of-concept Redis cache integration to make sure the API makes sense.

Since I've moved away from `Free` and a state-based interpreter, the new interpreter runs a `Fetch` instance into a `ConcurrentEffect`. 

There are some features that I need to add before finishing, namely:

- [x] `Monad[Fetch]#tailRecM`
- [x] Missing identities
- [x] Add caching
- [x] Test error handling
- [x] Test batching
- [x] Syntax
- [x] Update environment debug code
- [x] Update documentation
- [x] Update examples (blocked until http4s supports cats-effect >= 1.0.0)
- [x] Update README
- [x] Scala.js support
- [x] Coverage
- [ ] Bump version to `1.0.0-RC1`

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,13 +19,13 @@ For Scala 2.11.x and 2.12.x:
 [comment]: # (Start Replace)
 
 ```scala
-"com.47deg" %% "fetch" % "0.7.3"
+"com.47deg" %% "fetch" % "1.0.0-RC1"
 ```
 
 Or, if using Scala.js (0.6.x):
 
 ```scala
-"com.47deg" %%% "fetch" % "0.7.3"
+"com.47deg" %%% "fetch" % "1.0.0-RC1"
 ```
 
 [comment]: # (End Replace)
@@ -66,8 +66,6 @@ Returning `ConcurrentEffect` instances from the fetch methods allows us to speci
 We'll implement a dummy data source that can convert integers to strings. For convenience, we define a `fetchString` function that lifts identities (`Int` in our dummy data source) to a `Fetch`.
 
 ```scala
-import scala.concurrent.duration._
-
 import cats.data.NonEmptyList
 import cats.effect._
 import cats.temp.par._
@@ -125,9 +123,12 @@ def fetchOne[F[_] : ConcurrentEffect]: Fetch[F, String] =
 Let's run it and wait for the fetch to complete, we'll use `IO#unsafeRunTimed` for testing purposes, which will run an `IO[A]` to `Option[A]` and return `None` if it didn't complete in time:
 
 ```scala
+import scala.concurrent.duration._
+// import scala.concurrent.duration._
+
 Fetch.run[IO](fetchOne).unsafeRunTimed(5.seconds)
-// --> [68] One ToString 1
-// <-- [68] One ToString 1
+// --> [48] One ToString 1
+// <-- [48] One ToString 1
 // res0: Option[String] = Some(1)
 ```
 
@@ -146,8 +147,8 @@ When executing the above fetch, note how the three identities get batched and th
 
 ```scala
 Fetch.run[IO](fetchThree).unsafeRunTimed(5.seconds)
-// --> [69] Batch ToString NonEmptyList(1, 2, 3)
-// <-- [69] Batch ToString NonEmptyList(1, 2, 3)
+// --> [49] Batch ToString NonEmptyList(1, 2, 3)
+// <-- [49] Batch ToString NonEmptyList(1, 2, 3)
 // res1: Option[(String, String, String)] = Some((1,2,3))
 ```
 
@@ -179,12 +180,12 @@ When executing the above fetch, note how the three identities get requested in p
 
 ```scala
 Fetch.run[IO](fetchUnbatchedThree).unsafeRunTimed(5.seconds)
-// --> [69] One UnbatchedToString 2
-// --> [70] One UnbatchedToString 3
-// <-- [70] One UnbatchedToString 3
-// --> [71] One UnbatchedToString 1
-// <-- [69] One UnbatchedToString 2
-// <-- [71] One UnbatchedToString 1
+// --> [49] One UnbatchedToString 2
+// --> [48] One UnbatchedToString 3
+// --> [51] One UnbatchedToString 1
+// <-- [48] One UnbatchedToString 3
+// <-- [49] One UnbatchedToString 2
+// <-- [51] One UnbatchedToString 1
 // res2: Option[(String, String, String)] = Some((1,2,3))
 ```
 
@@ -223,10 +224,10 @@ Note how the two independent data fetches run in parallel, minimizing the latenc
 
 ```scala
 Fetch.run[IO](fetchMulti).unsafeRunTimed(5.seconds)
-// --> [70] One ToString 1
-// <-- [70] One ToString 1
-// --> [68] One Length one
-// <-- [68] One Length one
+// --> [48] One ToString 1
+// <-- [48] One ToString 1
+// --> [50] One Length one
+// <-- [50] One Length one
 // res3: Option[(String, Int)] = Some((1,3))
 ```
 
@@ -247,8 +248,8 @@ While running it, notice that the data source is only queried once. The next tim
 
 ```scala
 Fetch.run[IO](fetchTwice).unsafeRunTimed(5.seconds)
-// --> [71] One ToString 1
-// <-- [71] One ToString 1
+// --> [51] One ToString 1
+// <-- [51] One ToString 1
 // res4: Option[(String, String)] = Some((1,1))
 ```
 
--- a/docs/src/main/tut/index.md
+++ b/docs/src/main/tut/index.md
@@ -15,13 +15,13 @@ For Scala 2.11.x and 2.12.x:
 [comment]: # (Start Replace)
 
 ```scala
-"com.47deg" %% "fetch" % "0.7.3"
+"com.47deg" %% "fetch" % "1.0.0-RC1"
 ```
 
 Or, if using Scala.js (0.6.x):
 
 ```scala
-"com.47deg" %%% "fetch" % "0.7.3"
+"com.47deg" %%% "fetch" % "1.0.0-RC1"
 ```
 
 [comment]: # (End Replace)
@@ -69,8 +69,6 @@ Returning `ConcurrentEffect` instances from the fetch methods allows us to speci
 We'll implement a dummy data source that can convert integers to strings. For convenience, we define a `fetchString` function that lifts identities (`Int` in our dummy data source) to a `Fetch`.
 
 ```tut:silent
-import scala.concurrent.duration._
-
 import cats.data.NonEmptyList
 import cats.effect._
 import cats.temp.par._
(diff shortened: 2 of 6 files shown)


shared/src/main/scala/fetch.scala
```diff
@@ -17,207 +17,497 @@
...
+          case Blocked(br, cont) =>
+            Blocked(br, map(cont)(f))
+        }
+      } yield result.asInstanceOf[FetchResult[B]])
+
+    override def product[A, B](fa: Fetch[A], fb: Fetch[B]): Fetch[(A, B)] =
+      Unfetch(for {
+        a <- fa.run
+        b <- fb.run
+        result = (a, b) match {
```
fedefernandez: Can this be expressed with an `apply` product? If it's possible I think it makes more sense.
```scala
(fa.run, fb.run).tupled.map {
  case (Done(a), Done(b)) => Done((a, b))
  //...
```

shared/src/main/scala/cache.scala
```diff
@@ -16,55 +16,32 @@
...
+  def insert[I, A](i: I, ds: DataSource[I, A], v: A): IO[DataSourceCache]
 }
 
 /**
  * A cache that stores its elements in memory.
  */
-case class InMemoryCache(state: Map[DataSourceIdentity, Any]) extends DataSourceCache {
-  override def get[A](k: DataSourceIdentity): Option[A] =
-    state.get(k).asInstanceOf[Option[A]]
+case class InMemoryCache(state: Map[(String, Any), Any]) extends DataSourceCache {
```
fedefernandez: Maybe we could implement the `InMemoryCache` with `Ref`
```scala
case class InMemoryCache(state: Ref[IO, Map[(String, Any), Any]) extends DataSourceCache
```
But in this case, you would need to find the place to instantiate this cache as a side effect (`Ref.of[IO, Map[(String, Any), Any](Map.empty)` returns `IO`

fedefernandez: `(String, Any)` represents the DataSource name and the "key", right?

Can we define both with a [value classes](https://docs.scala-lang.org/overviews/core/value-classes.html)?
```scala
final class DataSourceName(val name: String) extends AnyVal
final class DataSourceKey(val key: Any) extends AnyVal
```

purrgrammer:
I think this PR is ready for review. It's not finished since I still need to add syntax and a few combinators in the `Fetch` companion object but I'd like to receive feedback of what I've done so far.

I'm not entirely happy with having to require the `cats.Parallel[IO, IO.Par]` implicit in `DataSource#batch` but the default implementation needs it to do `NonEmptyList#parTraverse`:

```scala
  def batch(ids: NonEmptyList[I])(
    implicit P: P[IO, IO.Par]
  ): IO[Map[I, A]] =
    ids.parTraverse(fetchOneWithId).map(_.collect { case Some(x) => x }.toMap)
```

shared/src/main/scala/fetch.scala
```diff
@@ -17,207 +17,467 @@
...
-sealed trait FetchQuery[I, A] extends FetchRequest {
-  def dataSource: DataSource[I, A]
-  def identities: NonEmptyList[I]
-}
+  // A query to a remote data source
+  sealed trait FetchQuery extends FetchRequest {
+    def dataSource[I, A]: DataSource[I, A]
+    def identities[I]: NonEmptyList[I]
+  }
+  case class FetchOne[I, A](id: I, ds: DataSource[I, A]) extends FetchQuery {
```
gatorcse: This definition seems dangerous. It says that the underlying datasource has fixed types for `I` and `A`, but then the definitions below allow callers to specify the types.

shared/src/main/scala/fetch.scala
```diff
@@ -17,207 +17,467 @@
...
-  override def identities: NonEmptyList[I]  = ids
-}
+  // In-progress request
+  case class BlockedRequest(request: FetchRequest, result: FetchStatus => IO[Unit])
 
-final case class Concurrent(queries: NonEmptyList[FetchQuery[Any, Any]])
-    extends FetchOp[InMemoryCache]
-    with FetchRequest
+  /* Combines the identities of two `FetchQuery` to the same data source. */
+  private def combineIdentities[I](x: FetchQuery, y: FetchQuery): NonEmptyList[I] = {
```
gatorcse: This seems like a dangerous method due to the reason stated above. We know nothing about the underlying `Datastore`, including their key and value types, yet we specify below that their identity types will be the same. If the caller does not enforce this guarantee, won't this blow up?

README.md
```diff
@@ -38,77 +38,99 @@ Or, if using Scala.js (0.6.x):
...
-    Query.sync({
-      println(s"[${Thread.currentThread.getId}] One ToString $id")
-      Option(id.toString)
-    })
+
+  override def fetch(id: Int): IO[Option[String]] = {
+    IO(println(s"--> [${Thread.currentThread.getId}] One ToString $id")) >>
+    IO.sleep(10.milliseconds) >>
+    IO(println(s"<-- [${Thread.currentThread.getId}] One ToString $id")) >>
+    IO(Option(id.toString))
```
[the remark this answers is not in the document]
purrgrammer: Good point, I'm going to update usages of `IO#apply` to be more explicit with `IO#delay` and `IO#pure`.

shared/src/main/scala/fetch.scala
```diff
@@ -17,207 +17,467 @@
...
+    def identities[I]: NonEmptyList[I]
+  }
+  case class FetchOne[I, A](id: I, ds: DataSource[I, A]) extends FetchQuery {
+    override def identities[I]: NonEmptyList[I] = NonEmptyList(id.asInstanceOf[I], List.empty[I])
+    override def dataSource[I, A]: DataSource[I, A] = ds.asInstanceOf[DataSource[I, A]]
+  }
+  case class Batch[I, A](ids: NonEmptyList[I], ds: DataSource[I, A]) extends FetchQuery {
+    override def identities[I]: NonEmptyList[I] = ids.toNonEmptyList.asInstanceOf[NonEmptyList[I]]
+    override def dataSource[I, A]: DataSource[I, A] = ds.asInstanceOf[DataSource[I, A]]
+  }
```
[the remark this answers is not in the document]
purrgrammer: Great point, those castings are there to work around type errors while I was developing, they are not needed.

shared/src/main/scala/fetch.scala
```diff
@@ -189,8 +189,12 @@ object `package` {
         }
       } yield result)
 
+    // todo: can be tail recursive?
     def tailRecM[A, B](a: A)(f: A => Fetch[Either[A, B]]): Fetch[B] =
-      ???
+      f(a).flatMap(_ match {
+        case Left(a) => tailRecM(a)(f)
+        case Right(b) => pure(b)
+      })
```
fedefernandez: it would be possible to express this with `fold` (I guess you'd need the `cats.syntax.either._`)?

```scala
f(a).flatMap(_.fold(tailRecM(_)(f), pure(_))
```

purrgrammer: I will update it. However, I'm not sure if we can write `tailRecM` as a tail recursive method since we need `.flatMap`.

README.md
```diff
@@ -123,39 +145,42 @@ val fetchThree: Fetch[(String, String, String)] = (fetchString(1), fetchString(2
...
   override def name = "Length"
-  
-  override def fetchOne(id: String): Query[Option[Int]] = {
-    Query.async((ok, fail) => {
-      println(s"[${Thread.currentThread.getId}] One Length $id")
-      ok(Option(id.size))
-    })
+
+  override def fetch(id: String): IO[Option[Int]] = {
+    IO.delay(println(s"--> [${Thread.currentThread.getId}] One Length $id")) >>
```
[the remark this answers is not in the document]
purrgrammer: Indeed, I changed to `IO#delay` to be more explicit. I'll use `ConcurrentEffect` when I add #156 changes to this branch.

docs/src/main/tut/index.md
```diff
@@ -54,41 +70,50 @@ Data Sources take two type parameters:
 
 ```scala
 import cats.data.NonEmptyList
+import cats.Parallel
 
 trait DataSource[Identity, Result]{
   def name: String
-  def fetchOne(id: Identity): Query[Option[Result]]
-  def fetchMany(ids: NonEmptyList[Identity]): Query[Map[Identity, Result]]
+  def fetch(id: Identity): IO[Option[Result]]
```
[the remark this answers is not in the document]
purrgrammer: Done in #156

shared/src/test/scala/FetchSyntaxTests.scala
```diff
@@ -17,44 +17,47 @@
...
-  import fetch.syntax._
   import TestHelper._
 
-  val ME = FetchMonadError[Future]
+  override val executionContext: ExecutionContext = ExecutionContext.Implicits.global
+  implicit val timer: Timer[IO] = IO.timer(executionContext)
+  implicit val cs: ContextShift[IO] = IO.contextShift(executionContext)
 
-  implicit override def executionContext = ExecutionContext.Implicits.global
+  implicit def ioToFuture[A](io: IO[A]): Future[A] = io.unsafeToFuture()
```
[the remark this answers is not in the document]
purrgrammer: I'll use that instead, good point 👍

purrgrammer:
All checks passed :shipit:

raulraja:
This is great @purrgrammer, outstanding work. Merge at will!

peterneyens:
Do we plan on bumping the version to 2?

purrgrammer:
I was planning to bump the version to 1.0.0 when we merge these changes, they are pretty substantial and deserve a major version change.

PR #156, opened 2018-09-07
POC: Redesign without `IO`

gatorcse:
I experimented with changing #155 to not require `IO`. The stated reasons for specifying `IO` were for `Ref` and `Deferred`, but those only require `Sync` and `Concurrent` respectively. I wanted to see if we could remove the requirement for IO, allowing Monix or other implementations to be used.

This is just a proof of concept. I didn't update the tests or anything, it also adds a `F[_]` parameter all over the place. This can probably be wrapped up in an abstract class or something to simplify the API. I just wanted to see if it would work.

### DIFF
--- a/shared/src/main/scala/fetch.scala
+++ b/shared/src/main/scala/fetch.scala
@@ -22,27 +22,30 @@ import scala.util.control.NoStackTrace
 import scala.concurrent.duration.MILLISECONDS
 
 import cats._
-import cats.instances.list._
+//import cats.instances.list._
+import cats.data._
+import cats.implicits._
 import cats.effect._
 import cats.effect.concurrent.{Ref, Deferred}
-import cats.syntax.all._
-import cats.data.NonEmptyList
+import cats.temp.par._
+//import cats.syntax.all._
+//import cats.data.NonEmptyList
 
 
 object `package` {
   // Fetch queries
-  sealed trait FetchRequest extends Product with Serializable
+  sealed trait FetchRequest[F[_]] extends Product with Serializable
 
   // A query to a remote data source
-  sealed trait FetchQuery[I, A] extends FetchRequest {
+  sealed trait FetchQuery[F[_], I, A] extends FetchRequest[F] {
     def dataSource: DataSource[I, A]
     def identities: NonEmptyList[I]
   }
-  case class FetchOne[I, A](id: I, ds: DataSource[I, A]) extends FetchQuery[I, A] {
+  case class FetchOne[F[_], I, A](id: I, ds: DataSource[I, A]) extends FetchQuery[F, I, A] {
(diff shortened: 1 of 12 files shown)


shared/src/main/scala/fetch.scala
```diff
@@ -213,163 +216,163 @@ object `package` {
...
 
     /**
       * Run a `Fetch`, the result in the `IO` monad.
       */
-    def run[A](
-      fa: Fetch[A],
-      cache: DataSourceCache = InMemoryCache.empty
+    def run[F[_]: Sync: Par, A](
+      fa: Fetch[F, A],
+      cache: DataSourceCache[F] = InMemoryCache.empty[F]
```
gatorcse: This doesn't compile because the call to `.empty` requires the implicit `Sync` instance, which is in the implicit argument list that follows, but that is not important to the concept.

shared/src/main/scala/fetch.scala
```diff
@@ -213,163 +216,163 @@ object `package` {
...
       c <- cache.get
     } yield (c, result)
 
-    private def performRun[A](
-      fa: Fetch[A],
-      cache: Ref[IO, DataSourceCache],
-      env: Option[Ref[IO, Env]]
+    private def performRun[F[_]: Sync: Par, A](
+      fa: Fetch[F, A],
+      cache: Ref[F, DataSourceCache[F]],
```
gatorcse: Doesn't need `IO`, just an implicit `Sync`

shared/src/main/scala/fetch.scala
```diff
@@ -213,163 +216,163 @@ object `package` {
...
       implicit
-        CS: ContextShift[IO]
-    ): Fetch[A] =
-      Unfetch(
+        CS: ContextShift[F]
+    ): Fetch[F, A] =
+      Unfetch[F, A](
         for {
-          deferred <- Deferred[IO, FetchStatus]
+          deferred <- Deferred[F, FetchStatus]
```
gatorcse: Doesn't need `IO`, just an implicit `Concurrent`

raulraja:
@gatorcse great stuff 👏 . The clear advantage of being more generic is that we can run Fetch to monix Task and any other data type that already provides instances for cats-effect.

purrgrammer:
I managed to make it compile, removing the `F` parameter from a couple places like `DataSource` (it is instead now passed to `DataSource#fetch` and `DataSource#batch`). I still need to update the examples/tests and the documentation, but I think I'm going to incorporate this into the `redesign` branch.

Awesome work @gatorcse, thanks for doing the heavy lifting with this changes.

purrgrammer:
Added these improvements to #155 ✨

PR #158, opened 2018-09-19
Fixes micrositePushWith Setting

juanpedromoreno:
This PR fixes the microsite auto-publish for the next releases. Also, it upgrades to the latest version of `sbt`.


purrgrammer:
✨

Issue #45, opened 2016-05-26
Identities with optional results


ghost:
The current semantics of Fetch are that identities **must** have a result. If a result is missing, the fetch execution short-circuits and fails. Maybe we should support the notion of optional fetches, much like Clump does [with the .optional method](http://getclump.io/#composition-behavior).

gregbeech:
The current behaviour is actually pretty surprising, that a method with this signature:

```scala
def fetchOne(id: I): Query[Option[A]]
```

Causes the whole execution to fail. It's not clear why the result is an `Option` if `None` isn't a valid result.

peterneyens:
@gregbeech I agree.

I was planning to rework `Query` a bit anyway after the next fetch release supporting Cats 1.0.0-MF is out in one of the next few days. I'll see if we can support optional results as well.

prurph:
:+1:

In particular this would make fetch an even better fit for GraphQL APIs, where the default is to make everything optional, so parts of a query can fail, while others succeed. For example, in the GitHub API if you search for foo and bar and only foo succeeds, you'll see the foo data and and an error at the bar node. This is very challenging to implement with fetch because `fetchMany` will immediately yield a `MissingIdentifiers` so you can only see which fetches failed/didn't return results.

paulpdaniels:
@peterneyens @raulraja Any movement on this? Cats has since stabilized. I'm curious what design changes you had in mind to support this feature.

purrgrammer:
> Causes the whole execution to fail. It's not clear why the result is an Option if None isn't a valid result.

The `None` signals that the identity hasn't been found, the current assumption of the library is that every identity exists, that's why `Fetch(identity)` returns a `Fetch[A]`. As the interest in this issue suggests we should support optional identities. 

For individual identities you could create an optional fetch doing `Fetch.optional(identity)` which returns a `Fetch[Option[A]]`, in case of optional identities when fetching a collection I'm not sure how to approach it, one option would be to add a version of `Fetch.multiple` that marks fetched identities as optional.

> @peterneyens @raulraja Any movement on this? Cats has since stabilized. I'm curious what design changes you had in mind to support this feature.

[...]

PR #159, opened 2018-09-19
Introduce Fetch#optional for performing optional fetches

purrgrammer:
You can create `Fetch` values with `Fetch#optional` instead of `Fetch#apply` and you'll get back a `Fetch[F, Option[A]]` instead of a `Fetch[F, A]`. 

I still want to experiment working with optional fetches using `OptionT` but this change solves #45 

 - [ ] Update docs
 - [ ] Add more testing

### DIFF
--- a/shared/src/main/scala/fetch.scala
+++ b/shared/src/main/scala/fetch.scala
@@ -241,6 +241,25 @@ object `package` {
         ))
       )
 
+    def optional[F[_] : ConcurrentEffect, I, A](id: I, ds: DataSource[I, A]): Fetch[F, Option[A]] =
+      Unfetch[F, Option[A]](
+        for {
+          deferred <- Deferred[F, FetchStatus]
+          request = FetchOne(id, ds)
+          result = deferred.complete _
+          blocked = BlockedRequest(request, result)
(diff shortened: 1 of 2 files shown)


shared/src/main/scala/fetch.scala
```diff
@@ -241,6 +241,25 @@ object `package` {
...
+          result = deferred.complete _
+          blocked = BlockedRequest(request, result)
+          anyDs = ds.asInstanceOf[DataSource[Any, Any]]
+          blockedRequest = RequestMap(Map(anyDs -> blocked))
+        } yield Blocked(blockedRequest, Unfetch[F, Option[A]](
+          deferred.get.map {
+            case FetchDone(a) =>
+              Done(Some(a)).asInstanceOf[FetchResult[F, Option[A]]]
+            case FetchMissing() =>
+              Done(None).asInstanceOf[FetchResult[F, Option[A]]]
```
peterneyens: I don't think we need the cast here. Can we try `Option.empty[A]`?

JorgeCastilloPrz:
Added to the ongoing redesign of Kollect.

PR #160, opened 2018-09-20
DataSourceCache parameterised to F[_]

purrgrammer:
I changed `DataSourceCache` so we can parameterise it to `F[_] : Monad`. It also helps for implementing custom caches that make use of io for serialization/deserialization.

### DIFF
--- a/docs/src/main/tut/docs.md
+++ b/docs/src/main/tut/docs.md
@@ -535,16 +535,12 @@ trait DataSourceCache[F[_]] {
 Let's implement a cache that forgets everything we store in it.
 
 ```tut:silent
-case class ForgetfulCache[F[_]]() extends DataSourceCache[F] {
-  def insert[I, A](i: I, v: A, ds: DataSource[I, A])(
-    implicit C: ConcurrentEffect[F], P: Par[F]
-  ): F[DataSourceCache[F]] =
-    C.pure(this)
+case class ForgetfulCache[F[_] : Monad]() extends DataSourceCache[F] {
+  def insert[I, A](i: I, v: A, ds: DataSource[I, A]): F[DataSourceCache[F]] =
+    Applicative[F].pure(this)
 
-  def lookup[I, A](i: I, ds: DataSource[I, A])(
-    implicit C: ConcurrentEffect[F], P: Par[F]
-  ): F[Option[A]] =
-    C.pure(None)
+  def lookup[I, A](i: I, ds: DataSource[I, A]): F[Option[A]] =
+    Applicative[F].pure(None)
 }
 
 def forgetfulCache[F[_] : ConcurrentEffect : Par] = ForgetfulCache[F]()
@@ -553,7 +549,7 @@ def forgetfulCache[F[_] : ConcurrentEffect : Par] = ForgetfulCache[F]()
 We can now use our implementation of the cache when running a fetch.
 
 ```tut:book
-def fetchSameTwice[F[_] : ConcurrentEffect]: Fetch[F, (User, User)] = for {
+def fetchSameTwice[F[_] : ConcurrentEffect : Par]: Fetch[F, (User, User)] = for {
   one <- getUser(1)
   another <- getUser(1)
 } yield (one, another)
--- a/docs/src/main/tut/docs.md
+++ b/docs/src/main/tut/docs.md
@@ -535,6 +535,8 @@ trait DataSourceCache[F[_]] {
 Let's implement a cache that forgets everything we store in it.
 
 ```tut:silent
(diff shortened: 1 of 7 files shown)


shared/src/main/scala/cache.scala
```diff
@@ -28,42 +30,49 @@ final class DataSourceResult(val result: Any) extends AnyVal
...
-trait DataSourceCache {
-  def lookup[F[_] : ConcurrentEffect, I, A](i: I, ds: DataSource[I, A]): F[Option[A]]
-  def insert[F[_] : ConcurrentEffect, I, A](i: I, v: A, ds: DataSource[I, A]): F[DataSourceCache]
-  def insertMany[F[_]: ConcurrentEffect, I, A](vs: Map[I, A], ds: DataSource[I, A]): F[DataSourceCache] =
-    vs.toList.foldLeftM(this)({
-      case (c, (i, v)) => c.insert(i, v, ds)
-    })
+trait DataSourceCache[F[_]] {
+  def lookup[I, A](i: I, ds: DataSource[I, A])(
+    implicit C: ConcurrentEffect[F], P: Par[F]
```
peterneyens: It would be nice if we didn't put these constraints on the methods anymore.

Now that we have parametrized `DataSourceCache` on `F`, the implementation could specify the constraints.
For example for `InMemoryCache` we would only need `Applicative` (`Monad` for `foldLeftM`).

purrgrammer: Good point, I'm going to try to remove these constraints.

purrgrammer:
@juanpedromoreno @peterneyens Updated with Peter's feedback, let me know if I can go ahead and merge this.

shared/src/main/scala/fetch.scala
```diff
@@ -268,17 +268,28 @@ object `package` {
     def run[F[_]]: FetchRunner[F] = new FetchRunner[F]
 
     private[fetch] class FetchRunner[F[_]](private val dummy: Boolean = true) extends AnyVal {
+      def apply[A](
+        fa: Fetch[F, A]
+      )(
+        implicit
+          P: Par[F],
+          C: ConcurrentEffect[F],
```
juanpedromoreno: Could be this replaced by `Monad`?

shared/src/main/scala/fetch.scala
```diff
@@ -289,9 +300,20 @@ object `package` {
     def runEnv[F[_]]: FetchRunnerEnv[F] = new FetchRunnerEnv[F]
 
     private[fetch] class FetchRunnerEnv[F[_]](private val dummy: Boolean = true) extends AnyVal {
+      def apply[A](
+        fa: Fetch[F, A]
+      )(
+        implicit
+          P: Par[F],
+          C: ConcurrentEffect[F],
```
juanpedromoreno: Same here

docs/src/main/tut/docs.md
```diff
@@ -526,7 +526,7 @@ The default cache is implemented as an immutable in-memory map, but users are fr
 There is no need for the cache to be mutable since fetch executions run in an interpreter that uses the state monad. Note that the `update` method in the `DataSourceCache` trait yields a new, updated cache.
 
 ```scala
-trait DataSourceCache {
+trait DataSourceCache[F[_]] {
   def insert[F[_] : ConcurrentEffect, I, A](i: I, ds: DataSource[I, A], v: A): DataSourceIdentity, v: A): F[DataSourceCache]
   def lookup[F[_] : ConcurrentEffect, I, A](i: I, ds: DataSource[I, A]): F[Option[A]]
```
peterneyens: These are no longer correct, right?

shared/src/main/scala/fetch.scala
purrgrammer: I'm afraid is not. The effect type `F[_]` of the `DataSource` instances requires a `ConcurrentEffect[F]` and `Par[F]`. For running `F` we need `ContextShift[F]` and `Timer[F]`so these are the less restrictive implicits.

peterneyens: I think we need `Sync` for `Ref.of` and the `DataSource` methods still require `ConcurrentEffect` at the moment.
So `performRun` and co need `ConcurrentEffect` as well.

docs/src/main/tut/docs.md
purrgrammer: You are right!

purrgrammer:
@peterneyens @juanpedromoreno i opened #163 to discuss about the implicits required when implementing data sources, let me know your thoughts in there. 

Can I merge this :shipit:  ?

PR #162, opened 2018-09-25
GitHub example

purrgrammer:
A small example of using Fetch to request data from Github. In this case we are doing the same data fetches that [org](https://github.com/47deg/org) does. The main example is:

 - Given a Github org, we fetch their public repos
   - For each repo, we get their contributors
   - For each repo, we get their languages

Fetch executes the above in two rounds of execution, making as much requests in parallel as possible.

The second example simply fetches multiple repositories in parallel.

Under the hood we're also following Github pagination links.

### DIFF
--- a/examples/src/test/scala/GithubExample.scala
+++ b/examples/src/test/scala/GithubExample.scala
@@ -1,0 +1,281 @@
+/*
+ * Copyright 2016-2018 47 Degrees, LLC. <http://www.47deg.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
+
+import cats._
+import cats.data.NonEmptyList
+import cats.effect._
+import cats.instances.list._
+import cats.syntax.all._
+import cats.temp.par._
+
+import io.circe._
+import io.circe.generic.semiauto._
+
+import org.http4s._
+import org.http4s.headers._
+import org.http4s.util.CaseInsensitiveString
+import org.http4s.circe._
+import org.http4s.client._
+import org.http4s.client.dsl._
+import org.http4s.client.blaze._
+import org.scalatest.{Matchers, WordSpec}
+
+import fetch.{DataSource, Env, Fetch}
+
+class GithubExample extends WordSpec with Matchers {
+  implicit val executionContext = ExecutionContext.Implicits.global
+
+  val ACCESS_TOKEN: String = sys.env("ORG_GITHUB_TOKEN")
+
+  implicit val t: Timer[IO]         = IO.timer(executionContext)
+  implicit val cs: ContextShift[IO] = IO.contextShift(executionContext)
+
+  // http4s client which is used by the datasources
+
+  def client[F[_]: ConcurrentEffect] =
+    Http1Client[F](BlazeClientConfig.defaultConfig)
+
+  // Github HTTP api
+
+  val GITHUB: Uri = Uri.unsafeFromString("https://api.github.com")
+
+  val REL_NEXT = """ rel="next" """".trim
+
+  def hasNext[F[_]: ConcurrentEffect](res: Response[F]): Boolean =
+    res.headers.get(CaseInsensitiveString("link")).fold(false)(_.value.contains(REL_NEXT))
+
+  def getNextLink(raw: String): String =
+    raw.value.takeWhile(_ != ';').dropWhile(_ == '<').takeWhile(_ != '>').trim
+
+  def getNext[F[_]: ConcurrentEffect](res: Response[F]): F[Uri] =
+    res.headers
+      .get(CaseInsensitiveString("link"))
+      .fold(Sync[F].raiseError[Uri](new Exception("next not found")))(raw =>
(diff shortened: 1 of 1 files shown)


examples/src/test/scala/GithubExample.scala
```diff
@@ -0,0 +1,310 @@
...
+  def client[F[_]: ConcurrentEffect] =
+    Http1Client[F](BlazeClientConfig.defaultConfig)
+
+  // Github HTTP api
+
+  val GITHUB: Uri = Uri.unsafeFromString("https://api.github.com")
+
+  val REL_NEXT = "rel=\"next\"".r
+
+  def hasNext[F[_]: ConcurrentEffect](res: Response[F]): Boolean =
```
fedefernandez: This function is used only inside `private` functions, can we define it as `private` too?

examples/src/test/scala/GithubExample.scala
```diff
@@ -0,0 +1,310 @@
...
+    REL_NEXT
+      .findFirstMatchIn(raw)
+      .fold(
+        Sync[F].raiseError[String](new Exception("Couldn't find next link"))
+      )(m => {
+        Sync[F].pure(m.before.toString.split(",").last.trim.dropWhile(_ == '<').takeWhile(_ != '>'))
+      })
+  }
+
+  def getNext[F[_]: ConcurrentEffect](res: Response[F]): F[Uri] =
```
fedefernandez: same as above

examples/src/test/scala/GithubExample.scala
```diff
@@ -0,0 +1,310 @@
...
+  val REL_NEXT = "rel=\"next\"".r
+
+  def hasNext[F[_]: ConcurrentEffect](res: Response[F]): Boolean =
+    res.headers
+      .get(CaseInsensitiveString("Link"))
+      .fold(false)({ h =>
+        REL_NEXT.findFirstIn(h.value).isDefined
+      })
+
+  def getNextLink[F[_]: ConcurrentEffect](raw: String): F[String] = {
```
fedefernandez: same as above

examples/src/test/scala/GithubExample.scala
```diff
@@ -0,0 +1,310 @@
...
+      forks_count: Int,
+      stargazers_count: Int,
+      watchers_count: Int,
+      languages_url: String,
+      contributors_url: String
+  )
+
+  implicit val repoD: Decoder[Repo] = deriveDecoder
+
+  private def fetchOrgRepos[F[_]](c: Client[F], req: Request[F])(
```
fedefernandez: These three functions follow the same pattern, I'm guessing if we could just write one generic function and then move the `hasNext`, `getNextLink` and `getNext` inside of it:
```scala
private def fetchDataRecursively[F[_]: ConcurrentEffect, A](c: Client[F], req: Request[F])(decode: Response[F] => F[List[A]])(
      implicit C: ConcurrentEffect[F]
  ): F[List[A]] = {

    val REL_NEXT = "rel=\"next\"".r

    def hasNext(res: Response[F]): Boolean =
      res.headers
        .get(CaseInsensitiveString("Link"))
        .fold(false)({ h =>
          REL_NEXT.findFirstIn(h.value).isDefined
        })

    def getNextLink(raw: String): F[String] = {
      REL_NEXT
        .findFirstMatchIn(raw)
        .fold(
          Sync[F].raiseError[String](new Exception("Couldn't find next link"))
        )(m => {
          Sync[F].pure(m.before.toString.split(",").last.trim.dropWhile(_ == '<').takeWhile(_ != '>'))
        })
    }

    def getNext(res: Response[F]): F[Uri] =
      res.headers
        .get(CaseInsensitiveString("Link"))
        .fold(Sync[F].raiseError[Uri](new Exception("next not found")))(
          raw => getNextLink(raw.value).map(Uri.unsafeFromString(_))
        )

    for {
      result <- c.fetch[List[A]](req) {
        case Status.Ok(res) =>
          if (hasNext(res)) {
            for {
              data <- decode(res)
              nxt  <- getNext(res)
              newReq = req.withUri(nxt)
              moreData <- fetchDataRecursively(c, newReq)(decode)
            } yield data ++ moreData
          } else
            res.as[List[A]]
        case res =>
          C.raiseError(new Exception(res.body.toString))
      }
    } yield result
  }
```

fedefernandez: Then we could have:
```scala
private def fetchOrgRepos[F[_]](c: Client[F], req: Request[F])(
      implicit C: ConcurrentEffect[F]
  ): F[List[Repo]] = {
    implicit val reposED: EntityDecoder[F, List[Repo]] = jsonOf
    fetchDataRecursively[F, Repo](c, req)(_.as[List[Repo]])
  }

private def fetchLanguages[F[_]](c: Client[F], req: Request[F])(
      implicit C: ConcurrentEffect[F]
  ): F[List[Language]] = {
    implicit val objED: EntityDecoder[F, JsonObject] = jsonOf
    fetchDataRecursively[F, Language](c, req)(_.as[JsonObject].map(_.toList.map(_._1)))
  }

private def fetchContributors[F[_]](c: Client[F], req: Request[F])(
      implicit C: ConcurrentEffect[F]
  ): F[List[Contributor]] = {
    implicit val objED: EntityDecoder[F, List[Contributor]] = jsonOf
    fetchDataRecursively[F, Contributor](c, req)(_.as[List[Contributor]])
  }
```

examples/src/test/scala/GithubExample.scala
```diff
@@ -0,0 +1,310 @@
...
+import org.http4s.util.CaseInsensitiveString
+import org.http4s.circe._
+import org.http4s.client._
+import org.http4s.client.dsl._
+import org.http4s.client.blaze._
+import org.scalatest.{Matchers, WordSpec}
+
+import fetch.{DataSource, Env, Fetch}
+
+class GithubExample extends WordSpec with Matchers {
```
raulraja: Does it make sense to use http://47deg.github.io/github4s/ instead of the custom Http client and unsafe building of Uris?

purrgrammer: AFAIK it doesn't support the call to get languages but is something I can add to `github4s`. Good point.

PR #166, opened 2018-10-29
Remove Par typeclass

purrgrammer:
After discussing with @peterneyens how we could get rid of the `Par` typeclass he suggested using `Concurrent#start` and `Fiber`s to achieve parallelism whenever we used `parTraverse`. This is my first approach to implementing it.

- Removes `cats-par` dependency
- Removes `Par` implicit requirement 
- Implements parallelism in terms of `Concurrrent#start` and `Fiber`'s
- [x] Support cancellation of parallel fetches

### DIFF
--- a/shared/src/main/scala/execution.scala
+++ b/shared/src/main/scala/execution.scala
@@ -28,7 +28,7 @@ private object FetchExecution {
       fibers <- effects.traverse(CF.start(_))
       runFibers = fibers.traverse(_.join)
       _ <- CF.handleErrorWith(runFibers)(error => {
-        fibers.traverse(_.cancel *> CF.raiseError(error))
+        fibers.traverse(_.cancel) *> CF.raiseError(error)
       })
       results <- runFibers
     } yield results
--- a/shared/src/main/scala/execution.scala
+++ b/shared/src/main/scala/execution.scala
@@ -24,14 +24,8 @@ private object FetchExecution {
   def parallel[F[_], A](effects: NonEmptyList[F[A]])(
     implicit CF: ConcurrentEffect[F]
   ): F[NonEmptyList[A]] =
-    for {
-      fibers <- effects.traverse(CF.start(_))
-      runFibers = fibers.traverse(_.join)
-      _ <- CF.handleErrorWith(runFibers)(error => {
-        fibers.traverse(_.cancel) *> CF.raiseError(error)
-      })
-      results <- runFibers
-    } yield results
-
+    effects.traverse(CF.start(_)).flatMap(fibers =>
+      fibers.traverse(_.join).onError({ case _ => fibers.traverse_(_.cancel) })
+    )
 }
 
--- a/shared/src/main/scala/execution.scala
+++ b/shared/src/main/scala/execution.scala
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2016-2018 47 Degrees, LLC. <http://www.47deg.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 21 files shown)


purrgrammer:
I'm not sure if we can support running parallel computations and making them cancelable, since the `Concurrent#cancelable` looks like the following:


```scala
import cats.effect.{Async, Fiber, IO, CancelToken}

trait Concurrent[F[_]] extends Async[F] {
  // ...
  def cancelable[A](k: (Either[Throwable, A] => Unit) => CancelToken[F]): F[A]
}
```

For now I've made the fibers we spawn uncancelable, so in case the fetch computation is cancelled, it will terminate as soon as all concurrent fibers have finished executing.

purrgrammer:
I have some ideas for supporting cancellation of parallel computations, I'll update the PR when I manage to implement it.

raulraja:
Looks great so far, let me know when ready for final review. 🍻

purrgrammer:
@raulraja not sure why the coverage doesn't cover the newly introduced `execution.scala` file but this PR is ready for review!

shared/src/test/scala/FetchSpec.scala
```diff
@@ -0,0 +1,50 @@
...
+
+@DoNotDiscover
+class FetchSpec extends AsyncFreeSpec with Matchers {
+  override val executionContext: ExecutionContext = ExecutionContext.Implicits.global
+  implicit val timer: Timer[IO] = IO.timer(executionContext)
+  implicit val cs: ContextShift[IO] = IO.contextShift(executionContext)
+
+  def countFetches(r: Request): Int =
+    r.request match {
+      case FetchOne(_, _)       => 1
```
raulraja: Formatting seems off in a few match cases here. Are we enforcing auto formatting with Scalafmt or similar?

purrgrammer: Not right now, I can open a separate PR for adding scalafmt and not polluting this one

shared/src/main/scala/execution.scala
```diff
@@ -0,0 +1,37 @@
...
+import cats.syntax.all._
+
+private object FetchExecution {
+  def parallel[F[_], A](effects: NonEmptyList[F[A]])(
+    implicit CF: ConcurrentEffect[F]
+  ): F[NonEmptyList[A]] =
+    for {
+      fibers <- effects.traverse(CF.start(_))
+      runFibers = fibers.traverse(_.join)
+      _ <- CF.handleErrorWith(runFibers)(error => {
```
peterneyens: - We could use `onError { case _ => fibers.traverse(_.cancel) }` instead of `handleErrorWith` + `raiseError`.

- We can also use `traverse_` (or `nonEmptyTraverse_` given that we have a `NonEmptyList`).

- Are we running `runFibers` twice? First in `_ <- CF.handleErrorWIth(...)` and again in `results <- runFibers`?

Maybe
```scala
effects.nonEmptyTraverse(CF.start(_)).flatMap { fibers =>
  fibers
    .nonEmptyTraverse(_.join)
    .onError { case _ => fibers.nonEmptyTraverse_(_.cancel) }
}
```

peterneyens: `Fiber` also has an `Applicative` instance, so we could do something like :
```scala
effects.nonEmptyTraverse(Cf.start(_))(CF.compose[Fiber[F, ?]])
```

Not sure if that is better though.

shared/src/test/scala/FetchSpec.scala
```diff
@@ -0,0 +1,50 @@
...
+
+import org.scalatest.{AsyncFreeSpec, Matchers, DoNotDiscover}
+
+import scala.concurrent._
+import java.util.concurrent._
+import scala.concurrent.duration._
+
+import cats.effect._
+
+@DoNotDiscover
```
peterneyens: What does this annotation do?

purrgrammer: Since we are using `FetchSpec` as a base class for specifications we don't want the test framework to find it and run it (it has no tests, just the execution environment). The `DoNotDiscover` annotation tells the testing framework to don't consider the `FetchSpec` class as one of our specs so it doesn't run it.

purrgrammer:
Cleaned this up following @peterneyens suggestions, let me know if :heavy_check_mark:

jtjeferreira:
any chance to make a release (RC or milestone) with this PR included? Would like to try this for a small project...

PR #171, opened 2018-11-21
Add F[_] type parameter to DataSource

purrgrammer:
In the old version of Fetch, one needed to define the data source as an
implicit object. This way, we ensured that each DataSource had only one instance using reference equality.

A data source used to look like the following:

```scala
implicit object Users extends DataSource[UserId, User]{
  // ...
}
```

Now that we've introduced an `F[_]` type parameter in `DataSource` we no
longer can define it as an `object`. Thus, we need a way to identify
requests to a data source to be able to perform optimizations.

Now there's a new trait called `Data`, that identifies requests for a
particular piece of data, and is used to identify and group requests for
the same data.

A `Data` definition looks like the following:

```scala
object Users extends Data[Int, User] {
  def name = "Users"
}
```

Then, when defining the data source for users, we have to specify which
`Data` the `DataSource` is fetching:

```scala
def usersSource[F[_] : ConcurrentEffect]: DataSource[F, Int, User] = new DataSource[F, Int,
User] {
  def data = Users

  def CF = ConcurrentEffect[F]

  // ...
}
```

Now that we have `Data` and `DataSource` we can create a `Fetch` por a `User`:

```scala
def fetchUser[F[_] : ConcurrentEffect](id: Int): Fetch[F, User] =
  Fetch(id, usersSource)
```

See the `examples` directory and the tests for plenty of examples.

### DIFF
--- a/examples/src/test/scala/DoobieExample.scala
+++ b/examples/src/test/scala/DoobieExample.scala
@@ -77,19 +77,17 @@ object DatabaseExample {
   object Authors extends Data[AuthorId, Author] {
     def name = "Authors"
 
-    def db[F[_]]: DataSource[F, AuthorId, Author] =
+    def db[F[_]: ConcurrentEffect]: DataSource[F, AuthorId, Author] =
       new DataSource[F, AuthorId, Author] {
         def data = Authors
 
-        override def fetch(id: AuthorId)(
-            implicit C: ConcurrentEffect[F]
-        ): F[Option[Author]] =
+        override def CF = ConcurrentEffect[F]
+
+        override def fetch(id: AuthorId): F[Option[Author]] =
           Database.transactor
             .flatMap(Queries.fetchById(id).transact(_))
 
-        override def batch(ids: NonEmptyList[AuthorId])(
-            implicit C: ConcurrentEffect[F]
-        ): F[Map[AuthorId, Author]] =
+        override def batch(ids: NonEmptyList[AuthorId]): F[Map[AuthorId, Author]] =
           Database.transactor
             .flatMap(Queries.fetchByIds(ids).transact(_))
             .map { authors =>
--- a/examples/src/test/scala/DoobieExample.scala
+++ b/examples/src/test/scala/DoobieExample.scala
@@ -29,56 +29,13 @@ import scala.concurrent.{ExecutionContext}
 
 import fetch._
 
-class DoobieExample extends WordSpec with Matchers {
-  implicit val executionContext = ExecutionContext.Implicits.global
-
-  implicit val t: Timer[IO]         = IO.timer(executionContext)
-  implicit val cs: ContextShift[IO] = IO.contextShift(executionContext)
-
-  val createTransactor: IO[Transactor[IO]] =
-    H2Transactor[IO]("jdbc:h2:mem:test;DB_CLOSE_DELAY=-1", "sa", "")
-
+object DatabaseExample {
   case class AuthorId(id: Int)
   case class Author(id: Int, name: String)
 
-  val dropTable = sql"DROP TABLE IF EXISTS author".update.run
-
-  val createTable = sql"""
-     CREATE TABLE author (
-       id INTEGER PRIMARY KEY,
-       name VARCHAR(20) NOT NULL UNIQUE
-     )
-    """.update.run
-
-  def addAuthor(author: Author) =
-    sql"INSERT INTO author (id, name) VALUES(${author.id}, ${author.name})".update.run
-
-  val authors: List[Author] =
-    List("William Shakespeare", "Charles Dickens", "George Orwell").zipWithIndex.map {
-      case (name, id) => Author(id + 1, name)
-    }
-
-  val xa: Transactor[IO] =
-    (for {
-      xa <- createTransactor
-      _  <- (dropTable *> createTable *> authors.traverse(addAuthor)).transact(xa)
-    } yield xa).unsafeRunSync()
-
-  val authorDS = new DataSource[AuthorId, Author] {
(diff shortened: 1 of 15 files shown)


shared/src/main/scala/datasource.scala
```diff
@@ -25,21 +25,26 @@ import cats.syntax.all._
...
 
+  /** The identity of the data source.
+   */
+  def identity: DataSourceIdentity =
+    DataSourceIdentity(name)
+
   /** Fetch one identity, returning a None if it wasn't found.
    */
-  def fetch[F[_] : ConcurrentEffect](id: I): F[Option[A]]
+  def fetch(id: I)(implicit C: ConcurrentEffect[F]): F[Option[A]]
```
diesalbla: Do you need this implicit parameter on the method itself, or can it be delegated as a constructor parameter of the implementing classes?

If you leave it for the implementation, you can still have a tests mock classes where you do not need the restrictions, just the `Applicative` to use a `pure`.

purrgrammer: > Do you need this implicit parameter on the method itself, or can it be delegated as a constructor parameter of the implementing classes?

Yes, I want it to be available in the methods to use the `ConcurrentEffect` API and have that implicit in scope.

examples/src/test/scala/DoobieExample.scala
```diff
@@ -29,77 +29,86 @@ import scala.concurrent.{ExecutionContext}
...
-      sql"SELECT * FROM author WHERE id = $id".query[Author].option
+    def createTransactor[F[_]: ConcurrentEffect] =
+      H2Transactor[F]("jdbc:h2:mem:test;DB_CLOSE_DELAY=-1", "sa", "")
 
-    def fetchByIds(ids: NonEmptyList[AuthorId]): ConnectionIO[List[Author]] = {
-      val q = fr"SELECT * FROM author WHERE" ++ Fragments.in(fr"id", ids)
-      q.query[Author].list
-    }
+    def transactor[F[_]: ConcurrentEffect]: F[Transactor[F]] =
+      for {
```
diesalbla:
```suggestion
      createTransactor.flatTap { _.trans(dropTable *> createTable *> authors.traverse(addAuthor) )
```
The `flatTap` allows you to _tee_ and use the output for a side-effectful computation.

shared/src/main/scala/datasource.scala
diesalbla: Right, but the benefit of an `implicit` value, or type-class instance, is that you do not need to feed it in the super-class. You can let a particular subclass to require it when it uses it in the implementation. IMHO, we should lean on using the _abstract_ methods to just declare a dependency of input parameters (the `id`) to the output (the `Option[A]`).

shared/src/main/scala/datasource.scala
```diff
@@ -25,21 +25,26 @@ import cats.syntax.all._
...
 
   /** Fetch many identities, returning a mapping from identities to results. If an
    * identity wasn't found, it won't appear in the keys.
    */
-  def batch[F[_] : ConcurrentEffect](ids: NonEmptyList[I]): F[Map[I, A]] =
+  def batch(ids: NonEmptyList[I])(implicit C: ConcurrentEffect[F]): F[Map[I, A]] =
```
diesalbla: So, one problem with using `implicit` parameters is that you can not override them or remove them from a method in the subclass: the subclasses must keep a similar signature to the parent. 

We could leave here no implementation, or add one that only used an `Applicative` constraint (to use with `List.traverse`), and then move the implementation using `ConcurrentEffect` to a different mix-in trait. 

```Scala
trait ConcurrentBatch[F[_]] extends DataSource[F] {
   implicit protected ConcF: ConcurrentEffect[F] 
   def batch(ids: NonEmptyList[I]): F[Map[I, A]] = /****/
}
```

purrgrammer: The `ConcurrentEffect` type class is something we can't escape from since it'll be required by Fetch when running or creating fetches. Having the implicit `ConcurrentEffect` in scope is something that you want when defining data fetches, since very few of those will require only `Applicative`. 

Remember that Fetch is a library for reordering and optimizing IO, in particular reading data. In practice, you will be using the `ConcurrentEffect` API to express resource allocation, asynchronous effects and lifting IO computations to `F` so I think it's fine.

examples/src/test/scala/DoobieExample.scala
purrgrammer: I tried it but I got a type mismatch between `doobie.h2.H2Transactor` and `doobie.Transactor`.

shared/src/main/scala/fetch.scala
```diff
@@ -122,21 +118,30 @@ object `package` {
...
     RequestMap(
       x.m.foldLeft(y.m) {
-        case (acc, (ds, blocked)) => {
-          val combinedReq: BlockedRequest[F] = acc.get(ds).fold(blocked)(combineRequests(blocked, _))
-          acc.updated(ds, combinedReq)
+        case (acc, (dsId, (ds, blocked))) => {
+          val combined = acc.get(dsId).fold(
+            (ds, blocked)
+          )({
+            case (ds, req) => {
```
diesalbla: The `ds` variable here shadows the one from line 136. Should they always be the same?

purrgrammer: Yes.

diesalbla: In that case, perhaps we should use back-quotes? As it is right now, the second `ds` is a completely different variable, that just happens to have the same name.

JorgeCastilloPrz:
Can you open an issue on Kollect to replicate this in case it makes sense?

PR #173, opened 2019-01-08
Upgrade CI build

diesalbla:
Upgrade Ruby version to 2.3.0.
Add othe JDKs
Upgrade Scala version 2.12 to 2.12.8


purrgrammer:
It seems that for some builds sbt cannot find the Scalajs compiler version, see [this build log](https://travis-ci.org/47deg/fetch/jobs/476652681#L841).

PR #182, opened 2019-03-04
Fetch#liftIO

### LINKED ISSUE #181: Support for lifting IO actions without related Data/DataSource into Fetch
This is something that Omer Zach brought up in the Gitter channel, and we'd need to do some modifications to Fetch in order to support it. The idea is that, if you have an `IO[A]` or `IO[Option[A]]` that represents a Fetch but can't be batched, you should be able to lift it to Fetch. These IO actions wouldn't be batched or cached, so they execute every time a Fetch is run regardless of the contents of the cache.

```scala
def getUser: IO[User] = ???
def maybeGetUser: IO[Option[User]] = ???

def fetchUser[F[_] : ConcurrentEffect] : Fetch[F, User] = Fetch.liftIO(getUser)
def maybeFetchUser[F[_] : ConcurrentEffect] : Fetch[F, Option[User]] = Fetch.liftIO(maybeGetUser)
```

As a "side-effect", we'd be able to embed arbitrary IO actions in a Fetch, even those that perform writes/logging/debugging or anything else.

purrgrammer:
Closes #181

### DIFF
--- a/shared/src/main/scala/fetch.scala
+++ b/shared/src/main/scala/fetch.scala
@@ -307,6 +307,15 @@ object `package` {
         ))
       )
 
+    def liftIO[F[_] : ConcurrentEffect, A](io: IO[A]): Fetch[F, A] =
+      Unfetch[F, A](for {
+        either <- ConcurrentEffect[F].liftIO(io).attempt
+        result = either match {
+          case Left(err) => Throw[F, A](log => UnhandledException(err, log))
+          case Right(r) => Done[F, A](r)
+        }
+      } yield result)
+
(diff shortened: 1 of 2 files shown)


omerzach:
Anything I can do to help get this merged? We'd like to use it as soon as it's ready.

raulraja:
@omerzach This is now in master. @purrgrammer can we do a minor release or is there something else we need to wait for?

omerzach:
Thank you!

purrgrammer:
Should be available in 1.0.1-SNAPSHOT @omerzach

Issue #163, opened 2018-09-25
Less restrictive DataSource effect implicits


purrgrammer:
Currently the methods that fetch data in `DataSource` require an effect type `F` with `Par[F]` and `ConcurrentEffect[F]`.

`Par` is required due to the `DataSource#batch` implementation defaulting to running individual fetches in parallel. If we move away from it, Data sources that don't implement batching will run their fetches sequentially. `Par[F]` will still be required when creating or running a fetch to `F`, since is used when running multiple batches in parallel.

`ConcurrentEffect` is perhaps too restrictive, since we may not need the cancellation semantics that `Concurrent` provides. We could use `Effect` here, thoughts?

![cats-effect-typeclasses](https://typelevel.org/cats-effect/img/cats-effect-typeclasses.svg)

peterneyens:
I think with `Concurrent` we should be able to implement `batch` in parallel as well. The code won't end up looking as nice as with `Par`, but that is probably a price worth paying if we can loose the `Par` constraint in a lot of places.

fedefernandez:
It'll be possible to make the restriction to `Effect` and define a couple of implementations if there is a `ConcurrentEffect` instance?

ALPSMAC:
2 cents from the peanut gallery for what it's worth... 

From a users perspective moving from earlier versions of Fetch, it was very surprising to me to have to sprinkle implicit `Timer` and `ContextShift` magic into my otherwise `Monix.Task`-based DAOs - e.g.:

```scala
def fetchString[F[_] : ConcurrentEffect](n: Int): Fetch[F, String] =
    Fetch(n, ToStringSource)

//Why do I need to do all of this in order to interpret a Fetch into an IO?
//I understand *why* but it's counter-intuitive given that when we 
//run the Fetch to produce the IO we still haven't generated any side-effects.
val ec: ExecutionContext = implicitly[ExecutionContext]
implicit val timer: Timer[Task] = IO.timer(ec)
implicit val cs: ContextShift[IO] = IO.contextShift(ec)

val io: IO[String] = Fetch.run[IO](fetchString(123))

val task: Task[String] = Task.fromEffect(io)
```

So if I've got code like this:

```scala
trait KVStore{
   def get(key: String): Task[String]
}
```

[...]

PR #184, opened 2019-03-13
Loosen cats-effect constraints

kubukoz:
Touches #163 a bit. ConcurrentEffect replaced with Concurrent, ContextShift removed.

Let me know what you think about: replacing Timer usage with Clock (a Clock is implicitly available if a Timer is)

### DIFF
--- a/examples/src/test/scala/DoobieExample.scala
+++ b/examples/src/test/scala/DoobieExample.scala
@@ -64,12 +64,12 @@ object DatabaseExample {
         case (name, id) => Author(id + 1, name)
       }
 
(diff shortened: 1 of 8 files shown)


purrgrammer:
First of all thanks for taking the time to improve Fetch, is greatly appreciated. I've pushed a branch called `loosen-constraints` with your commits so it's build by Travis with our config, it should be good to merge when tests are green :green_heart:  

> Let me know what you think about: replacing Timer usage with Clock (a Clock is implicitly available if a Timer is)

:+1:

kubukoz:
Superseded by #187

PR #187, opened 2019-03-15
Loosen implicit constraints

purrgrammer:
Rebasing the commits in #184 on top of master so the test run with the updated Travis configuration.


purrgrammer:
I think we can merge this and close #184. All credit goes to @kubukoz by the way, you rock!

PR #189, opened 2019-04-25
Lift Concurrent actions to Fetch

### LINKED ISSUE #188: Support for lifting `F[_]: Concurrent` without related Data/DataSource into Fetch
This is a request for somewhat generalization of the issue brought up in https://github.com/47deg/fetch/issues/181
that was mentioned in the Gitter channel

as Fetch by itself requires `Concurrent` it'd be helpful to have a method like
```scala
def liftF[F[_]: Concurrent](ft: F[T]): Fetch[F, T]
```
to not be restricted for optimized flow for only `IO` monad

purrgrammer:
Closes #188

### DIFF
--- a/shared/src/main/scala/fetch.scala
+++ b/shared/src/main/scala/fetch.scala
@@ -311,22 +311,20 @@ object `package` {
       )
 
     def liftIO[F[_] : ConcurrentEffect, A](io: IO[A]): Fetch[F, A] =
-      Unfetch[F, A](for {
-        either <- ConcurrentEffect[F].liftIO(io).attempt
-        result = either match {
+      Unfetch[F, A](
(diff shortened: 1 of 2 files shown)


shared/src/main/scala/fetch.scala
```diff
@@ -316,6 +319,16 @@ object `package` {
         }
       } yield result)
 
+    def liftF[F[_] : Concurrent, A](f: F[A]): Fetch[F, A] =
+      Unfetch[F, A](for {
+        either <- f.attempt
```
diesalbla: I prefer avoiding single-line for comprehensions: 
```Scala
Unfetch[F, A] ( f.attempt.map {
  case Left(err) => Throw[F, A](log => UnhandledException(err, log))
  case Right(r)  => Done[F, A](r)
})
```

PR #191, opened 2019-04-26
Remove Unfetch wrapper class.

diesalbla:
This commit removes the Unfetch case class, a wrapper for a
F[FetchResult], and reduces the `Fetch` trait into a mere type alias.

The `Unfetch` class, as defined, is a mere artifice used to provide
an instance of the `Monad` type-class for a `F[FetchResult]`.

Instead of creating a wrapper class, and a Monad instance for that
wrapper class, in this PR we simply create a more specific monad
instance for an `F[FetchResult]`, one which handles both the `F` and
the inner `FetchResult` ADT structure.

### DIFF
--- a/shared/src/main/scala/fetch.scala
+++ b/shared/src/main/scala/fetch.scala
@@ -157,98 +157,78 @@ object `package` {
   private[fetch] final case class Throw[F[_], A](e: Log => FetchException) extends FetchResult[F, A]
 
   // Fetch data type
-
-  sealed trait Fetch[F[_], A] {
-    private[fetch] def run: F[FetchResult[F, A]]
-  }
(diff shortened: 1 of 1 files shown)


peterneyens:
I guess this would require a fetch 2.0 release? Is is worth it?

purrgrammer:
> I guess this would require a fetch 2.0 release? Is is worth it?

Since this would break API compatibility it'd require a 2.0.0 release. While I agree that Unfetch is not needed and is a mere wrapper that implements the Fetch trait, don't think its worth breaking API compatibility just after the 1.0.0 release.

There are also some failing test in the Github example.

purrgrammer:
Closing for now.

PR #194, opened 2019-05-09
Add Fetch#runAll

purrgrammer:


shared/src/main/scala/fetch.scala
```diff
@@ -416,6 +416,36 @@ object `package` {
       } yield (c, result)
     }
 
+    /**
+      * Run a `Fetch` getting the log, cache and result in the `F` monad.
+      */
+    def runAll[F[_]]: FetchRunnerAll[F] = new FetchRunnerAll[F]
+
+    private[fetch] class FetchRunnerAll[F[_]](private val dummy: Boolean = true) extends AnyVal {
```
juanpedromoreno: What's the purpose of `dummy`?

purrgrammer: It's just a dummy attribute to be able to define a class with the apply method diferent arities.

diesalbla: What would be the problem of using just two functions? 

```Scala
def runAll[F[_]: Concurrent: Timer, A](fa: Fetch[F, A]): F[(Log, DataCache[F], A)] = 
  runAll(fa, InMemoryCache.empty[F])

def runAll[F[_]: Concurrent: Timer, A](fa: Fetch[F, A], cache: DataCache[F])
  : F[(Log, DataCache[F], A)] = /**/ 
```

purrgrammer: We use this technique to be able to partially apply the type parameters in `Fetch.run*` methods, so you can do `Fetch.run[IO](f)` and omit the second type parameter. It doesn't work with an overloaded method.

diesalbla: What about using a syntax extension on the `Fetch[F, A]` element?

```Scala
implicit class FetchRunOps[F[_], A](val fa: Fetch[F, A]){
  def runAll()(implicit /****/) = 

  def runAll(cache: DataCache[F])(implicit /***/) = 
}

purrgrammer: Still haven't added syntax for run* methods, it's actually a bit tricky. Opened https://github.com/47deg/fetch/issues/195 to address it.

PR #197, opened 2019-05-23
Improve Redis example

purrgrammer:


examples/src/test/scala/JedisExample.scala
```diff
@@ -58,83 +60,129 @@ object DataSources {
...
-    })({ in =>
-      Sync[F].delay(in.close())
+  def serialize[F[_], A](obj: A)(
+      implicit S: Sync[F]
+  ): F[ByteArray] = {
+    type Resources = (ObjectOutputStream, ByteArrayOutputStream)
+
+    def resources[F[_]](implicit S: Sync[F]): Resource[F, Resources] =
+      for {
+        byte <- byteOutputStream
```
diesalbla: We can extract a `flatProduct` function, of the type `F[A] => (A => F[B]) => F[(A, B)]`. Maybe there is something like that already.

examples/src/test/scala/JedisExample.scala
```diff
@@ -58,83 +60,129 @@ object DataSources {
...
       _ <- set(identity(i, data), s)
     } yield this
+
+  override def bulkInsert[I, A](vs: List[(I, A)], data: Data[I, A])(
+      implicit M: Monad[F]
+  ): F[DataCache[F]] =
+    for {
+      bin <- vs.traverse({
+        case (id, v) =>
+          for {
```
diesalbla: SIngle-line `for` comprehension is a `map`.

examples/src/test/scala/JedisExample.scala
purrgrammer: Found it: https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/FlatMap.scala#L114 Good catch!

examples/src/test/scala/JedisExample.scala
```diff
@@ -58,83 +60,129 @@ object DataSources {
...
+            bin <- Binary.serialize(v)
+          } yield (identity(id, data), bin)
```
diesalbla: There is an `identity` function in the scala standard library. Perhaps the function could be renamed to avoid confussion.

examples/src/test/scala/JedisExample.scala
```diff
@@ -58,83 +60,129 @@ object DataSources {
...
-    Sync[F].pure(binaryClient)
+  def connection: Resource[F, Jedis] =
+    Resource.fromAutoCloseable(Sync[F].delay(pool.getResource))
 
   private def get(i: Array[Byte]): F[Option[Array[Byte]]] =
-    Sync[F].bracket(client)({ c =>
-      Sync[F].pure(Option(c.get(i)))
-    })({ in =>
-      Sync[F].pure(in.close())
+    connection.use({ c =>
```
diesalbla:
```suggestion
    connection.use(c => Sync[F].delay(Option(c.get(i))) )
```
You can save the braces.

examples/src/test/scala/JedisExample.scala
```diff
@@ -58,83 +60,129 @@ object DataSources {
...
+        pipe.sync
+      }))
+
   private def identity[I, A](i: I, data: Data[I, A]): Array[Byte] =
     Binary.fromString(s"${data.identity} ${i}")
 
-  def lookup[I, A](i: I, data: Data[I, A]): F[Option[A]] =
-    get(identity(i, data)) >>= { case (raw) => Binary.deserialize(raw) }
+  override def lookup[I, A](i: I, data: Data[I, A]): F[Option[A]] =
+    for {
```
diesalbla: As a matter of personal preference, here I would use a `flatMap` with the pattern-match:
```Scala
get(identity(i, data)) >>= {
  case None    => Sync[F].pure(None)
  case Some(r) => Binary.deserialize[F, A](r)
}
```

examples/src/test/scala/JedisExample.scala
```diff
@@ -58,83 +60,108 @@ object DataSources {
...
+        pipe.sync
+      }))
+
+  private def cacheId[I, A](i: I, data: Data[I, A]): Array[Byte] =
     Binary.fromString(s"${data.identity} ${i}")
 
-  def lookup[I, A](i: I, data: Data[I, A]): F[Option[A]] =
-    get(identity(i, data)) >>= { case (raw) => Binary.deserialize(raw) }
+  override def lookup[I, A](i: I, data: Data[I, A]): F[Option[A]] =
+    get(cacheId(i, data)).flatMap(_ match {
```
diesalbla: You do not need the `_ match`. Just the two cases between braces make up the literal of a `PartialFunction`, which (regrettably) is also a `Function`

PR #212, opened 2019-09-16
chore(dependencies): Update cats and monix dependencies.

### LINKED ISSUE #211: Update Cats 2.0.0 and Monix 3.0.0
Both Cats and Monix recently had major version bumps, cats-effect is at 2.0.0 and monix is at 3.0.0 (though I believe the latter is only for the example docs). We are looking to upgrade our code base but would also need fetch to update as well. 

I can submit a PR to that effect.

paulpdaniels:
Updates cats and monix to their latest versions, as well as other cats derived things like atto and http4s.

Closes #211

### DIFF
--- a/project/ProjectPlugin.scala
+++ b/project/ProjectPlugin.scala
@@ -21,7 +21,7 @@ object ProjectPlugin extends AutoPlugin {
     lazy val commonCrossDependencies =
       Seq(
 		libraryDependencies ++=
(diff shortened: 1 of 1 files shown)


purrgrammer:
Thanks a lot @paulpdaniels, I've opened a PR with your commit and a version bump in https://github.com/47deg/fetch/pull/214 I'll close this one after I merge the aforementioned PR.

PR #219, opened 2019-09-19
Update to Scala 2.13, SBT 1.3 and bump some libraries

bijancn:
WIP on updating to scala 2.13

Only do Ypartial for older scala versions

Update circe and doobie for example

Fix examples in tut


purrgrammer:
Awesome, will merge after Travis gives the green light. Thanks for your time and patience @bijancn !

bijancn:
No problem. Thank you for the very nice library! @purrgrammer

PR #227, opened 2019-10-25
Stack safety of combineRequests

abestel:
**Context**
Our application crashed because of a `StackOverflowError` in `combineRequests`.
We noticed it because we run the application with `-Xss256k`.

**Proposal**
Trampoline `combineRequests` and make it stack safe.

**Side note**
During my tests, I noticed that a big batch (5000 elements) was extremely slow to be processed.
Most of the time is spent into `combineIdentities` (half in `NonEmptyList.exists` and the other half in `NonEmptyList.:+`).
By changing `FetchQuery` to return identities as a `Set[I]`, the code is **much** faster (the big batch test went from minutes to seconds).
Any objection?

### DIFF
--- a/shared/src/main/scala/fetch.scala
+++ b/shared/src/main/scala/fetch.scala
@@ -59,7 +59,7 @@ object `package` {
 
   // In-progress request
 
-  private[fetch] final case class BlockedRequest[F[_]](request: FetchRequest, result: FetchStatus => F[Unit])
+  private[fetch] final case class BlockedRequest[F[_]](request: FetchRequest, result: CombinationTailRec[F])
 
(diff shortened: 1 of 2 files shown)


purrgrammer:
> During my tests, I noticed that a big batch (5000 elements) was extremely slow to be processed.
Most of the time is spent into combineIdentities (half in NonEmptyList.exists and the other half in NonEmptyList.:+).

`NonEmptyList` is not the right data structure to use internally, I was aware this could be a problem so thanks for actually fixing it!

> By changing FetchQuery to return identities as a Set[I], the code is much faster (the big batch test went from minutes to seconds).

:rocket: 

The changes look good, I'll merge and cut a new release. Thanks a lot for your time!

PR #229, opened 2019-10-25
Update Scala versions

purrgrammer:


project/ProjectPlugin.scala
```diff
@@ -125,7 +125,7 @@ object ProjectPlugin extends AutoPlugin {
         orgUpdateDocFilesSetting += baseDirectory.value / "tut",
         scalaOrganization := "org.scala-lang",
         scalaVersion := "2.13.0",
-        crossScalaVersions := List("2.11.12", "2.12.9", "2.13.0"),
+        crossScalaVersions := List("2.11.12", "2.12.10", "2.13.1"),
```
juanpedromoreno: We need to update [the travis file](https://github.com/47deg/fetch/blob/master/.travis.yml#L6-L7), otherwise, TravisCI will skip the build

PR #422, opened 2021-01-14
Fix compiler warnings across project

sloshy:
Some code used features of Scala or http4s that either were deprecated, or confused the compiler while being technically usable. This tweaks the code that's there just enough to get it all working without warnings.

* No more procedure syntax
* Pattern matching casts are now done explicitly
* http4s client `.fetch` syntax is deprecated, replaced w/ `.run.use`
* Replace `?` with `*` for kind projector usage


rafaparadela (approved):
Not sure about the explicit casting, but as long as removing compilation warnings is the motivation, I'm ok with that :-)
Thanks for the efforts.

Issue #306, opened 2020-05-07
Caching in readme


DenisNovac:
There is this example of caching at the end of readme:

```scala
def fetchTwice[F[_] : Concurrent]: Fetch[F, (String, String)] = for {
  one <- fetchString(1)
  two <- fetchString(1)
} yield (one, two)
```

Isn't this just a deduplication? I mean, results are not saved to the next call:

```scala
val l: Fetch[IO, (Option[String], Option[String])] = for {
    d1 <- data.fetchElem(0)
    d2 <- data.fetchElem(0)
  } yield (d1, d2)
  Fetch.run(l).unsafeRunSync()  // asks for elem 0 once
  Fetch.run(data.fetchElem(0)).unsafeRunSync()  // asks for elem 0 too, not cached?
```

sloshy:
As mentioned in the comment in the referenced issue, this is because a new cache is created each time you `run` a Fetch, by default. If, instead, creating a cache yourself that you can pass to each `Fetch` request. It will then use that and properly fetch the element from the local cache.

I will create a PR to clear this up in the README so that this behavior is much more obvious.

DenisNovac:
@sloshy that is a great addition, thank you.

PR #423, opened 2021-01-14
Expand Caching section in README

sloshy:
To address #306 which expressed some confusion on how caching and deduplication works, I've expanded the section to contain an example of sharing the cache between successive computations.

A couple other small changes that are tagging along:
* README should now contain the proper year range (unsure if this will work)
* Fixed some deprecated usage of `filterKeys` in the microsite docs


rafaparadela (approved):
Thanks for expanding such a great explanation

PR #443, opened 2021-02-19
Cross-building with Scala 3

benderpremier:
Closes https://github.com/47degrees/fetch/issues/439

Draft PR to cross build for scala 3
- Adapt the build for scala 3
- Use type pattern `:` instead of extractor `@`
- Update example dependencies

### Caveats:

- [ ] `fetchJS` tests do not compile for scala 3 as scalatest isn't published for scalaJS with scala 3 ( https://mvnrepository.com/artifact/org.scalatest/scalatest )
- [ ] `fetch-example` http4s isn't published for scala 3 yet.
- [ ] `fetch-example` monix isn't published for scala 3 yet.
- [ ] `sbt-mdoc-toc` sbt plugin isn't published for scala 3 yet. I think `sbt-mdoc-toc` transitively depends on `mdoc` being  ready for scala3. According to [this issue](https://github.com/scalameta/mdoc/issues/414), it is currently published for scala3 but only to allow worksheet support
- [ ] `sbt-scoverage` isn't published for scala 3 yet. (See [this issue](https://github.com/scoverage/sbt-scoverage/issues/316)


build.sbt
```diff
@@ -11,20 +11,32 @@ skip in publish := true
 lazy val fetch = crossProject(JSPlatform, JVMPlatform)
   .crossType(CrossType.Pure)
   .settings(commonCrossDependencies)
+  .settings(
+    scalaVersion := "3.0.0-M3",
+    crossScalaVersions := Seq("2.12.11", "2.13.2", "3.0.0-M3")
```
franciscodr, line 16: I guess we could use the latest versions for Scala 2.12 and 2.13:
- 2.12.13
- 2.13.5

benderpremier:
I've updated the 2 scala versions. I've tried to find workarounds to get the build to pass, but haven't settled on anything satisfactory. It looks like at this stage we still have too many dependencies that aren't published for scala 3.

juanpedromoreno:
Hi @benderpremier , thanks for helping with the Scala 3 migration! Some comments:

* I think scalatest is cross-compiled with [scala 3-RC1]
  * https://github.com/scalatest/scalatest/blob/d7b65ba585106ba655a856601d5e43cea965ab25/project/DottyBuild.scala
  * https://github.com/scalatest/scalatest/releases/tag/release-3.2.4
* `sbt-mdoc-toc`: I don't think we are using this plugin?
* Examples: as we are not publishing them in maven central, maybe we could avoid cross-compilation for example modules. Makes sense?
* sbt-scoverage: this one might be tricky. Perhaps, we could disable code-coverage for now.

Again, thanks!

benderpremier:
Thanks for the feedback @juanpedromoreno.
> I think scalatest is cross-compiled with [scala 3-RC1] 

It is cross compiled for scala 3 on the JVM, but not for scala 3 on scalaJS. For the moment I've disabled cross compilation on scalaJS with.
```scala
lazy val fetchJS = fetch.js
  .disablePlugins(ScoverageSbtPlugin)
  .settings(crossScalaVersions := scala2Versions)
```
> sbt-mdoc-toc: I don't think we are using this plugin?

I think we do [here](https://raw.githubusercontent.com/47degrees/fetch/db771e094b42d13866779645d245b9eeccb2490d/docs/README.md) (on line 9). It generates the table of content that we see in the `readme.md` at the root of the repo. However, I think we can scope the `mdoc` task to only run on `2.13` 

> Examples: as we are not publishing them in maven central, maybe we could avoid cross-compilation for example modules. Makes sense? 

Yes I think it makes sense. I'll only compile this project with `2.13`.

> sbt-scoverage: this one might be tricky. Perhaps, we could disable code-coverage for now.

If everyone is happy with that let's do it 💪.

benderpremier:
I've removed the code coverage plugins and adapted the build. I think if we want to publish for `3.0.0-RC1` that could do the trick.

rolandtritsch:
@benderpremier @franciscodr Some Scala3 porting projects will trigger some (significant) code-refactoring. It would be good to have this. Can we merge this and publish a new version (1.4.3)? Question: Am I right to assume that `sbt-scoverage` does not need to be updated, but that I need to set `coverageScalacPluginVersion` to `1.4.3` to make this work?

docs/README.md
```diff
@@ -1,6 +1,6 @@
...
+[![Join the chat at https://gitter.im/47deg/fetch](https://badges.gitter.im/47deg/fetch.svg)](https://gitter.im/47deg/fetch?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge) (http://codecov.io/github/47deg/fetch?branch=master) [![Maven Central](https://img.shields.io/badge/maven%20central-1.2.1-green.svg)](https://oss.sonatype.org/#nexus-search;gav~com.47deg~fetch*) [![License](https://img.shields.io/badge/license-Apache%202-blue.svg)](https://raw.githubusercontent.com/47deg/fetch/master/LICENSE) [![Latest version](https://img.shields.io/badge/fetch-1.2.1-green.svg)](https://index.scala-lang.org/47deg/fetch) [![Scala.js](http://scala-js.org/assets/badges/scalajs-0.6.15.svg)](http://scala-js.org) [![GitHub Issues](https://img.shields.io/github/issues/47deg/fetch.svg)](https://github.com/47deg/fetch/issues)
```
cb372, line 3: You removed half of the codecov badge but left `(http://codecov.io/github/47deg/fetch?branch=master)` - is that intentional?

fetch-examples/src/test/scala/GithubExample.scala
```diff
@@ -1,278 +1,278 @@
...
-          } else
-            res.as[List[A]]
-        case res =>
-          CF.raiseError(new Exception(s"Couldn't complete request, returned status: ${res.status}"))
-      }
-    } yield result
-  }
-
-}
+///*
```
cb372, line 1: What's the plan for this file? Should we delete it instead of commenting it out?

fetch/src/main/scala/fetch.scala
```diff
@@ -156,7 +156,10 @@ object `package` {
       y: BlockedRequest[F]
   ): BlockedRequest[F] =
     (x.request, y.request) match {
-      case (a @ FetchOne(aId, ds), b @ FetchOne(anotherId, _)) =>
+      case (a: FetchOne[Any, Any], b: FetchOne[Any, Any]) =>
```
cb372, line 159: This is a bit unfortunate. I can't claim to fully understand all this code, but as far as I can tell, it was taking advantage of bugs in the Scala compiler's GADT support in order to compile?

I had a quick look but I have a feeling that any attempt to address this properly would become a rewrite of most of the library. We could at least try to suppress the erasure warnings?

pepegar (commented):
This looks great Damian,

My only concern is the pattern matches in which we're matching patterns using the `pat: Type` form instead of capturing the match in a varialbe with `v @ Type`, we loose type safety that way.

Also, for HTTP4s, I think you can use one of the milestonre releases like `1.0.0-M1` (https://mvnrepository.com/artifact/org.http4s/http4s-dsl_2.13/1.0.0-M19) that crossbuilds for scala 3.0.0-RC1

fetch/src/main/scala/fetch.scala
pepegar: Something we can try if the `@` operator doesn't work is to rewrap the pattern again if needed?

```scala
      case (FetchOne(aId, ds), FetchOne(anotherId, ds2)) =>
        val a = FetchOne(aId, ds)
        val b = FetchOne(anotherId, ds2)
```

rolandtritsch:
Thanks for all the reviews!!! (cc: @benderpremier @franciscodr)

docs/README.md
benderpremier: That wasn't intentional. I'll fix it 👍 .

rolandtritsch:
@benderpremier I am interested to get this merged/published. Let me know, if/how I can help.

fetch/src/main/scala/fetch.scala
benderpremier: I tried that but I am getting the same compilation error:
```
[error] -- [E007] Type Mismatch Error: /Users/damien/code/fetch/fetch/src/main/scala/fetch.scala:167:48
[error] 167 |          val combined   = combineIdentities(a, b)
[error]     |                                                ^
[error]     |                      Found:    (b : fetch.package.FetchOne[I$2, A$2])
[error]     |                      Required: fetch.FetchQuery[I$1, A$1]
```
The compiler wants to prove that the type parameters of `FetchOne` a and b are the same but it can't. We lose those parameters when we use FetchRequest.

fetch-examples/src/test/scala/GithubExample.scala
benderpremier: I have uncommented all examples and only cross-compile them for scala2 versions. I assume we can wait until all libraries we depend on in the examples are published for scala3 to cross-compile the examples to scala3 as well.

fetch/src/main/scala/fetch.scala
benderpremier: I'll try to see if we can keep track of the `I` and `A` type params along the way. I gave it a quick go now and it doesn't seem so straightforward.

benderpremier:
I am not sure how we want to move forward with this. I agree that having the `a: FetchOne[Any, Any],` isn't great. As Chris pointed out, looks like we were taking advantage of a limitation of the scala2 compiler. I guess our choice are:
- merge with those `Any` which seems to be on par with what the library is doing now
- fix this before publishing for scala3

fetch/src/main/scala/fetch.scala
benderpremier: I spent some time this afternoon trying to keep track of the `I` and avoid the `FetchOne[Any, Any]`. However, I end up bubbling that type parameter all the way up in the API (in the pure and apply method of the `Fetch` object) which I don't think we want to do.

I am not sure there is a way around using `Any` here. If you look at how the `Fetch` are created, we widen to `Any` in  the fetch constructor methods (`apply`, `optional`) and then cast with `asInstanceOf[A]`. Looks like this is where we loose the `I` we would need here.

Also the monad instance for fetch, `implicit def fetchM[F[_]: Monad]: Monad[Fetch[F, *]]` would have to be reworked if we push the `I` all the way up which seems complicated.

juanpedromoreno (approved):
I'd suggest opening an issue for the `FetchOne[Any, Any]` improvement, and try to move this forward.

LGTM, thanks @benderpremier !

juanpedromoreno (dismissed):

rolandtritsch:
@benderpremier I agree with @juanpedromoreno:  Would suggest to merge now and create a followup ticket to work on `FetchOne`. WDYT?

benderpremier:
@juanpedromoreno I had to update to scala3 RC2. After merging master into this branch, scalatest got updated to 3.2.7 which is only published for RC2. Could you give it a ✅ ?

cb372 (approved):

PR #504, opened 2021-07-05
Add support for batching requests across unrelated fetches

jordiolivares:
I noticed that when processing Fetch requests across `Fetch.run` calls on an HTTP server the fetch requests weren't batched across executions. This means that if we were processing for example 3000 requests simultaneously where each Fetch is for a single ID, we would have 3000 calls to the `DataSource.fetch` method.

This PR thus adds support for bundling the calls within a given interval in order to produce a single `DataSource.batch` call instead. This would massively help throughput of HTTP servers at the cost of latency


jordiolivares:
For some reason the check fails due to an HTTP error even before the code is compiled/tested with this:

`HttpError: Resource not accessible by integration`

I have no clue why this is failing

juanpedromoreno:
In the meantime, I'm taking a look at the CI/CD issue with the PR labeler.

jordiolivares:
Seems like I've been in Scala 2.13 for so long I had forgotten about Scala 2.12, I've made the compilation fix, so feel free to review it again!

juanpedromoreno:
Thank so much @jordiolivares ! Great contribution 🙌

juanpedromoreno (approved):

PR #538, opened 2021-09-23
Fix problem with fibers being cancelled due to timeout and causing fetch deadlock

jordiolivares:
Hi again, it seems that the process for batching across requests suffers from a deadlock issue.

In particular it can be triggered if the timeout occurs between the time the queue has been popped and the buffer is updated.

To fix this I just wrapped it in an `F.uncancelable` to fix the deadlock, safely updating the reference before cancellation is finished

### DIFF
--- a/fetch/src/main/scala/datasource.scala
+++ b/fetch/src/main/scala/datasource.scala
@@ -114,7 +114,7 @@ object DataSource {
   ): Resource[F, DataSource[F, I, A]] = {
     type Callback = Either[Throwable, Option[A]] => Unit
     for {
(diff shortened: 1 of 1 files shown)


jordiolivares:
Seems like the check is failing only due to the GithubExample test having bad credentials, as per the test error:

```
java.lang.Exception: Couldn't complete request, returned status: 401 Unauthorized: Body:
[info] {"message":"Bad credentials","documentation_url":"https://docs.github.com/rest"}
```

juanpedromoreno (approved):
Thanks again @jordiolivares !

PR #590, opened 2022-01-24
Replace implicit traverse behavior with explicit batching

### LINKED ISSUE #585: Fetch breaks on Cats 2.7.0 due to Traverse implementation detail
Noticed that CI for upgrading to Cats Effect 3.3.x has been failing recently and after looking into it, it seems the reason it fails is due to calls to `.traverse` in tests not working as before. This is in tests that use `.traverse` in some way, such as assuming it will auto-batch requests.

### Root cause of problem
We're detecting batches based on calls to `map2`, `product`, and `productR` in our `Monad[Fetch]` implementation to optimize for batches, but now it seems as of Cats 2.7.0 (transitive dependency of Cats Effect 3.3.x) `map2` is not guaranteed to be called. This can be demonstrated by changing the version of Cats and noticing that the same tests fail as in the current CI pipelines for Cats Effect upgrade PRs.

### More details

In our tests we have a helper function `one` that fetches some pure value. If we have a test that fetches, in order, in a for-comprehension the following values as-written:

* `one(1)`
* `one(2)`
* `one(3)`
* `List(1, 2, 3).traverse(one(_))`

On Cats Effect 3.2.x it correctly batches the `List(1, 2, 3)` calls together under-the-hood. However, starting with Cats Effect 3.3.x it is now based on Cats 2.7 which appears to have changed the underlying implementation of `.traverse`.

To test this I added print-debugging statements to the `Monad[Fetch]` instance to see what methods were being called. Batch-detection is only done when certain methods are called such a `map2` or `product`, which `traverse` had historically used before. Now, it appears it is only calling `map` and `flatMap` instead.

Here is a sample of my print debugs from a test that is running on Cats 2.6.x:
```
Called map2 with:
blocked RequestMap(Map(16 -> (fetch.TestHelper$One$$anon$1@1e,BlockedRequest(FetchOne(2,fetch.TestHelper$One$@10),CombinationLeaf(<function1>))))) and blocked RequestMap(Map(16 -> (fetch.TestHelper$One$$anon$1@1d,BlockedRequest(FetchOne(3,fetch.TestHelper$One$@10),CombinationLeaf(<function1>)))))

combining...

Called map2 with:
blocked RequestMap(Map(16 -> (fetch.TestHelper$One$$anon$1@1f,BlockedRequest(FetchOne(1,fetch.TestHelper$One$@10),CombinationLeaf(<function1>))))) and blocked RequestMap(Map(16 -> (fetch.TestHelper$One$$anon$1@1d,BlockedRequest(Batch(NonEmptyList(2, 3),fetch.TestHelper$One$@10),CombinationSuspend(<function1>)))))

combining...

Called map with:
blocked RequestMap(Map(16 -> (fetch.TestHelper$One$$anon$1@1d,BlockedRequest(Batch(NonEmptyList(1, 2, 3),fetch.TestHelper$One$@10),CombinationSuspend(<function1>)))))

Called map with:
blocked RequestMap(Map(16 -> (fetch.TestHelper$One$$anon$1@1d,BlockedRequest(Batch(NonEmptyList(1, 2, 3),fetch.TestHelper$One$@10),CombinationSuspend(<function1>)))))

Called flatMap with:
blocked RequestMap(Map(16 -> (fetch.TestHelper$One$$anon$1@1d,BlockedRequest(Batch(NonEmptyList(1, 2, 3),fetch.TestHelper$One$@10),CombinationSuspend(<function1>)))))

Called map with:
value 3

Called map2 with:
values 2 and List(3)

Called map2 with:
values 1 and List(2, 3)

Called map with:
value List(1, 2, 3)
```

This is how it is resolving a `Fetch[F, List[Int]]` under-the-hood. To break it down, first it is combining the fetch for 2 and 3, into a fetch for `NonEmptyList(2, 3)`, and then it is combining that with the fetch for 1, which turns it into a `Batch(NonEmptyList(1,  2, 3))`. It's then able to run these batches and get the final list as expected, but only making one fetch for these three values.

On Cats 2.7.0, however, `map2` is never called. Instead, it appear to only ever call `map` and `flatMap`. Since `map2` or `product` or `productR` are the only methods that can detect and optimize for a batch like this, the tests that depend on this implementation detail of `traverse` will fail.

### Thoughts
This makes sense to me given my understanding of the laws of the type classes we are depending on. There is nothing about `.traverse` that says it has to be done using `.map2`. We should look into determining what should be batched some other way if possible, or representing the intent in a more obvious way.

Given `List(a, b, c).traverse(f)` should, in my mind, be equivalent to `f(a).flatMap(fa => f(b).flatMap(fb => f(c).map(fc => List(fa, fb, fc))))` it does not seem readily apparent how to extract from that the intent to batch without asking the user, anyway. Though I might be missing something.

sloshy:
Fixes #585 

This adds a new function to the `Fetch` companion object called `batchAll`. It takes a varargs of `Fetch[F, A]` and returns a `Fetch[F, List[A]]` back, with the assumption that all of the fetches provided are to be batched or run concurrently. This is to replace the previous behavior of automatically batching fetches that are included in `traverse` calls, because that functionality depended on implementation details of the Cats traverse instances, which have changed as of Cats 2.7.0.

Each of the tests that failed have been updated to now use `Fetch.batchAll` and test cases that specified the auto-batching behavior of `traverse` have been altered, and duplicated with changes, to demonstrate that this behavior no longer occurs.

Marked as WIP because I still have to finish the documentation but the code is ready for review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,10 +21,10 @@ A library for Simple & Efficient data access in Scala and Scala.js
 
 Add the following dependency to your project's build file.
 
-For Scala 2.11.x and 2.12.x:
+For Scala 2.11.x through 3.x:
 
 ```scala
-"com.47deg" %% "fetch" % "2.1.1"
+"com.47deg" %% "fetch" % "3.0.0"
 ```
 
 Or, if using Scala.js (0.6.x):
--- a/README.md
+++ b/README.md
@@ -27,10 +27,10 @@ For Scala 2.12.x through 3.x:
 "com.47deg" %% "fetch" % "3.0.0"
 ```
 
-Or, if using Scala.js (0.6.x):
+Or, if using Scala.js (1.0.x):
 
 ```scala
-"com.47deg" %%% "fetch" % "2.1.1"
+"com.47deg" %%% "fetch" % "3.0.0"
 ```
 
 
@@ -105,19 +105,14 @@ def fetchString[F[_] : Async](n: Int): Fetch[F, String] =
 
 ## Creating a runtime
 
-Since `Fetch` relies on `Concurrent` from the `cats-effect` library, we'll need a runtime for executing our effects. We'll be using `IO` from `cats-effect` to run fetches, but you can use any type that has a `Concurrent` instance.
+Since we'll use `IO` from the `cats-effect` library to execute our fetches, we'll need an `IORuntime` for executing our `IO` instances.
 
-For executing `IO`, we need a `ContextShift[IO]` used for running `IO` instances and a `Timer[IO]` that is used for scheduling. Let's go ahead and create them. We'll use a `java.util.concurrent.ScheduledThreadPoolExecutor` with a couple of threads to run our fetches.
-
-```scala
-import java.util.concurrent._
-import scala.concurrent.ExecutionContext
-
-val executor = new ScheduledThreadPoolExecutor(4)
-val executionContext: ExecutionContext = ExecutionContext.fromExecutor(executor)
-
-import cats.effect.unsafe.implicits.global
+```scala mdoc:silent
+import cats.effect.unsafe.implicits.global //Gives us an IORuntime in places it is normally not provided
 ```
+
+Normally, in your applications, this is provided by `IOApp`, and you should not need to import this except in limited scenarios such as test environments that do not have Cats Effect integration.
+For more information, and particularly on why you would usually not want to make one of these yourself, [see this post by Daniel Spiewak](https://github.com/typelevel/cats-effect/discussions/1562#discussioncomment-254838)
 
 ## Creating and running a fetch
 
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ A library for Simple & Efficient data access in Scala and Scala.js
 
 Add the following dependency to your project's build file.
 
-For Scala 2.11.x through 3.x:
+For Scala 2.12.x through 3.x:
 
 ```scala
 "com.47deg" %% "fetch" % "3.0.0"
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ For Scala 2.12.x through 3.x:
 "com.47deg" %% "fetch" % "3.0.0"
 ```
 
-Or, if using Scala.js (1.0.x):
+Or, if using Scala.js (1.8.x):
 
 ```scala
 "com.47deg" %%% "fetch" % "3.0.0"
--- a/fetch-examples/src/test/scala/Http4sExample.scala
+++ b/fetch-examples/src/test/scala/Http4sExample.scala
@@ -114,7 +114,7 @@ object HttpExample {
     fetchUserById(UserId(id))
 
   def fetchManyUsers[F[_]: Async](ids: List[Int]): Fetch[F, List[User]] =
-    ids.traverse(i => fetchUserById(UserId(i)))
+    Fetch.batchAll(ids.map(i => fetchUserById(UserId(i))): _*)
 
   def fetchPosts[F[_]: Async](user: User): Fetch[F, (User, List[Post])] =
     fetchPostsForUser(user.id).map(posts => (user, posts))
@@ -149,7 +149,7 @@ class Http4sExample extends AnyWordSpec with Matchers {
     def fetch[F[_]: Async]: Fetch[F, List[(User, List[Post])]] =
       for {
         users          <- fetchManyUsers(List(1, 2))
-        usersWithPosts <- users.traverse(fetchPosts[F])
+        usersWithPosts <- Fetch.batchAll(users.map(fetchPosts[F]): _*)
       } yield usersWithPosts
 
     val io = Fetch.runLog[IO](fetch)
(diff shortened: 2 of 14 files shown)


sloshy:
Tests compile and pass locally for me, so I'm not sure what that failure is about. Says the syntax isn't available but it's part of Cats 2.7.x anyway. Maybe it needs an `sbt clean`?

fetch-examples/src/test/scala/Http4sExample.scala
```diff
@@ -149,7 +149,7 @@ class Http4sExample extends AnyWordSpec with Matchers {
     def fetch[F[_]: Async]: Fetch[F, List[(User, List[Post])]] =
       for {
         users          <- fetchManyUsers(List(1, 2))
-        usersWithPosts <- users.traverse(fetchPosts[F])
+        usersWithPosts <- Fetch.batchAll(users.map(fetchPosts[F]): _*)
```
cb372, line 152: What do you think about providing some syntax sugar on `Seq[Fetch[F[_], A]]` allowing users to write this as `users.map(fetchPosts[F]).batchAll`?

fetch/src/main/scala/fetch.scala
```diff
@@ -360,6 +361,27 @@ object `package` {
...
+     * This should be used in code that previously relied on the auto-batching behavior of calling
+     * `traverse` on lists of `Fetch` values.
+     */
+    def batchAll[F[_]: Monad, A](fetches: Fetch[F, A]*): Fetch[F, List[A]] = {
+      fetches.toNeSeq
+        .map { nes =>
+          nes
+            .map(_.map(Chain.one(_)))
+            .reduceLeft { (fa, fb) =>
+              val mon = Monad[Fetch[F, *]]
```
cb372, line 377: super-nitpick, but is there any reason not to inline this into the line below?

cb372 (commented):
Looks good!

fedefernandez:
@sloshy it seems to be failing only for Scala `2.12.15`

```
value toNeSeq is not a member of fetch.package.Fetch[F,A]*
```

fetch/src/main/scala/fetch.scala
sloshy: Probably just an oversight on my part while I was trying to get things to compile. Metals was being weird at the time.

sloshy: Inlined :+1:

fetch-examples/src/test/scala/Http4sExample.scala
sloshy: Added `.batchAll` and `.batchAllWith` to mimic `.sequence` and `.traverse` respectively for batches

sloshy:
Okay, should be good now. The docs took a little while to figure out (sorry about that) but I updated the type classes in them as well to be more accurate, as well as included information about `batchAll` and the primary difference of this new version.

Since it's a breaking change I also assumed this would bump the version to 3.0.0. Does this sound alright?

README.md
```diff
@@ -21,10 +21,10 @@ A library for Simple & Efficient data access in Scala and Scala.js
...
 
-For Scala 2.11.x and 2.12.x:
+For Scala 2.12.x through 3.x:
 
 ```scala
-"com.47deg" %% "fetch" % "2.1.1"
+"com.47deg" %% "fetch" % "3.0.0"
 ```
 
 Or, if using Scala.js (0.6.x):
```
fedefernandez, line 30: We would need to update the version here too.

fedefernandez (approved):
Looks good!

README.md
cb372: Specifically, the Scala.js version is now 1.x (I assume) and the fetch version is 3.0.0

microsite/docs/docs.md
```diff
@@ -232,28 +235,23 @@ object OnlyBatched extends Data[Int, Int]{
...
-import cats.effect._
-import java.util.concurrent._
-import scala.concurrent.ExecutionContext
-import scala.concurrent.duration._
-
-val executor = new ScheduledThreadPoolExecutor(4)
-val executionContext: ExecutionContext = ExecutionContext.fromExecutor(executor)
-
-import cats.effect.unsafe.implicits.global
+import cats.effect.unsafe.implicits.global //Give
```
cb372, line 241: I think the `//Give` is a stray copy-paste or something?

cb372 (dismissed):
LGTM

microsite/docs/docs.md
sloshy: Ah I meant to say something there, but left it unfinished. Thanks.

README.md
sloshy: It's technically 1.8.x as per the compatibility rules here: https://www.scala-js.org/news/2021/12/10/announcing-scalajs-1.8.0/

> It is not forward binary compatible with 1.7.x: libraries compiled with 1.8.0 cannot be used with 1.7.x or earlier.

So I've adjusted it to match.

cb372 (approved):

fedefernandez (approved):
neat

PR #594, opened 2022-01-29
Add map2Eval override

sloshy:
As per #590 we're replacing implicit batching on versions `>=3.0.0` with explicit batching. This should still be supported on 2.x and support for batching breaking on Cats 2.7.0 should be considered a bug in need of fixing. In light of this, this PR:

* Bumps Cats Effect to the latest supported version, which depends on Cats 2.7.0 transitively
* Adds a map2Eval override so that implicit batching via traverse/sequence is still supported.


fedefernandez (approved):

cb372 (approved):
I didn't realise there was a 3-line fix for this. That changes everything! 😄

sloshy:
Okay in light of everything I've been looking into over the weekend, plus what I've developed recently in terms of experiments with alternate encodings of Fetch, I'm moving towards making this just go directly into the 3.x branch and ditch the idea of releasing another 2.x release. I've modified the documentation to account for this discrepancy as well.

fedefernandez (approved):

cb372 (approved):

PR #706, opened 2023-03-16
Logging cached requests

a-khakimov:
We suggest logging cached requests. These logs allow us to invalidate the cache more granularly. Additionally, it is helpful to see which requests have been cached in the logs.

```
Fetch execution 🕛 0,08 seconds

  [Round 1] 🕛 0,02 seconds
    [Batch] From `Users` with ids List(1, 2) cached false 🕛 0,02 seconds
  [Round 2] 🕛 0,00 seconds
    [Fetch one] From `Users` with id 2 cached true 🕛 0,00 seconds
  [Round 3] 🕛 0,01 seconds
    [Fetch one] From `Users` with id 4 cached false 🕛 0,01 seconds
  [Round 4] 🕛 0,01 seconds
    [Batch] From `Posts` with ids List(1, 2, 3) cached false 🕛 0,01 seconds
    [Batch] From `Users` with ids List(3) cached false 🕛 0,01 seconds
    [Batch] From `Users` with ids List(1, 2) cached true 🕛 0,00 seconds
```

### DIFF
--- a/fetch/src/main/scala/fetch.scala
+++ b/fetch/src/main/scala/fetch.scala
@@ -22,20 +22,21 @@ import cats.effect._
 import cats.effect.implicits._
 import cats.syntax.all._
 
-import scala.collection.immutable.Map
 import scala.util.control.NoStackTrace
 
 object `package` {
-  private[fetch] sealed trait FetchRequest extends Product with Serializable
+  private[fetch] sealed trait FetchRequest extends Product with Serializable {
+    def cached: Boolean
+  }
 
   private[fetch] sealed trait FetchQuery[I, A] extends FetchRequest {
     def data: Data[I, A]
     def identities: Set[I]
   }
-  private[fetch] final case class FetchOne[I, A](id: I, data: Data[I, A]) extends FetchQuery[I, A] {
+  private[fetch] final case class FetchOne[I, A](id: I, data: Data[I, A], cached: Boolean = false) extends FetchQuery[I, A] {
     override def identities: Set[I] = Set(id)
   }
-  private[fetch] final case class Batch[I, A](ids: NonEmptyList[I], data: Data[I, A])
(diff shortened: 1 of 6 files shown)


fetch/src/main/scala/fetch.scala
```diff
@@ -657,16 +658,18 @@ object `package` {
       T: Clock[F]
   ): F[List[Request]] =
     for {
+      startTime   <- T.monotonic.map(_.toMillis)
```
fedefernandez, line 661: Would it be possible to pass the flag here to avoid unnecessary calls to `Clock`? It will be minimal, but in a library that aims to be performant, every ms counts 😄

fedefernandez:
Thanks for the PR, @a-khakimov. Looks good. I left a comment and approved the CI build

fetch/src/main/scala/fetch.scala
a-khakimov: To be honest, I'm not sure. At first, I thought that for cached requests, the parameters "start" and "end" could be initialized with a constant, but it won't work because they are used to determine the first and last requests.

a-khakimov:
Pipeline failed due to the following case
```
[info] GithubExample:
[info] - We can fetch org repos *** FAILED ***
[info]   java.lang.Exception: Couldn't complete request, returned status: 401 Unauthorized: Body:
[info] {"message":"Bad credentials","documentation_url":"https://docs.github.com/rest"}
[info]   at GithubExample.$anonfun$fetchCollectionRecursively$17(GithubExample.scala:277)
[info]   at flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:163)
[info]   at flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:163)
...
```

But I doubt it's because of the current edits in the PR

fedefernandez (approved):
LGTM, thanks

